PR #1, opened 2020-10-21
Initial code

stephenworsley:
Copies most of the code from https://github.com/SciTools-incubator/iris-ugrid/blob/master/iris_ugrid/regrid.py and adds a skeleton structure for an iris regridding scheme. Note that the cope in `esmf_regridder.py` still lacks features we would want in an eventual iris regridder (e.g. handling masked values). The aim of this PR is to provide a base to begin building a proper iris ESMF regridder.


bjlittle:
@stephenworsley Awesome, looks like a decent starting point 💯 

LGTM 👍 

Over to you @zklaus 😉

PR #8, opened 2020-11-06
add cache refresh

bjlittle:
This PR adds two levels of control to force a refresh of `cirrus-ci` task caches:
1. Incorporate the `date` year `+%Y`, and week number `+%U` as part of a cache `fingerprint`, and using the `CACHE_AGE` on the week number to partition the week number into like week periods - for automated refreshes
1. Use the `X_CACHE_BUILD` number as part of a cache `fingerprint` to force a cache upload - for manual refreshes

The expected typical use case for bumping the cache build number would be to force a cache upload within a `CACHE_AGE` period. The build number should then be reset after the `CACHE_AGE` period has elapsed.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -13,7 +13,15 @@ container:
   image: python:3.8
 
 env:
-  PIP_CACHE_VERSION: "1"
+  # Maximum cache age in weeks before forcing a new cache upload.
+  CACHE_AGE: "2"
+  # Increment the build number to force new conda cache upload.
+  CONDA_CACHE_BUILD: "0"
+  # Increment the build number to force new nox cache upload.
+  NOX_CACHE_BUILD: "0"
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -13,7 +13,15 @@ container:
   image: python:3.8
 
 env:
-  PIP_CACHE_VERSION: "1"
+  # Maximum cache age in weeks before forcing a new cache upload.
+  CACHE_AGE: "2"
```
stephenworsley, line 17: I think this variable would be more accurately described as `CACHE_PERIOD`, as it is, `CACHE_AGE` implies that this describes how old the cache is.

stephenworsley (changes requested):
Just a minor issue with a variable name, otherwise this looks good.

stephenworsley (approved):

PR #9, opened 2020-11-06
pytest and codecov

bjlittle:
This PR adds `pytest` configuration and enables `codecov` test coverage metrics.

It also restructures the `esmf_regrid.test` directory structure, and also adds a common `esmf_regrid.tests.get_result_path` function.

The `flake8` plugin for `import-order` is also now correctly configured so that `esmf_regrid` is registered as the local package.


stephenworsley (approved):
Looks good.

PR #14, opened 2020-11-09
check nox conda requirements

bjlittle:
This PR streamlines the `nox` session for `tests`, taking advantage of re-using any associated cached `conda` environment by not always installing the environment requirements.

Rather, it first detects whether `pytest` is available, and if not installs the associated `conda` environment.


stephenworsley (approved):
Looks good.

PR #16, opened 2020-11-10
Update stale.yml

bjlittle:
Use the latest version of stale, see https://github.com/actions/stale/releases


stephenworsley (approved):
Looks good.

Issue #20, opened 2020-11-12
Support masked arrays


stephenworsley:
## ✨ Feature Request
Provide support for regridders to handle masked arrays. Regridders will allow control over the masked data tolerance, similar to iris area weighted conservative regridders.

## Motivation
Data on iris cubes is commonly masked. Furthermore, the regridding of masked data is a situation where the user may want more control over the regridding process for performance reasons. The handling of masked data is central to the design and motivation for this package.

stephenworsley:
It looks like there are two ways to normalise weights to account for masked values. These are described [here](http://www.earthsystemmodeling.org/esmf_releases/public/ESMF_8_0_1/esmpy_doc/html/NormType.html) and are discused [here](https://github.com/JiaweiZhuang/xESMF/issues/22#issuecomment-403172093). ESMF names these normalisation types as DSTAREA and FRACAREA. For DSTAREA, the weight associated with a source cell and a target cell is the proportion of the area of the **destination cell** which is taken up by the overlap of the source cell and the destination cell. FRACAREA is calculated similarly, except instead of weights being a proportion of the whole destination cell, they are instead a proportion of the **overlap of all unmasked source cells and that destination cell**, this means that the sum of all weights associated with a destination cell is always 1. It is interesting to note that for global grids, regridding with a DSTAREA norm is the equivalent of regridding with a FRACAREA norm but replacing all masked data with 0.

It is worth noting that both norm types have advantages. DSTAREA preserves the global mean value, but distorts values near masked points. FRACAREA is more accurate near masked points but ceases to conserve global mean.

Another point worth noting is that it is possible to derive FRACAREA weights from DSTAREA weights by normalising them, but it is not possible to derive DSTAREA weights from FRACAREA weights. For this reason, we extract the DSTAREA norm from ESMF which should allow us to calculate both.

PR #22, opened 2020-11-16
Add support for masked data

stephenworsley:
This addresses #20. It allows the regridding of masked arrays. It also supports a choice of normalisation options (as described [here](http://earthsystemmodeling.org/esmf_releases/public/ESMF_8_0_1/esmpy_doc/html/NormType.html)). This PR also improves the test coverage for the regridder which was missing in the initial code.

### DIFF
--- a/esmf_regrid/esmf_regridder.py
+++ b/esmf_regrid/esmf_regridder.py
@@ -290,8 +290,6 @@ class Regridder:
             A numpy array whose shape is compatible with self.tgt.
 
         """
-        # TODO implement masked array handling similar to other iris regridders.
-
         # A rudimentary filter is applied to mask data which is mapped from an
         # insufficiently large source. This currently only accounts for discrepancies
         # between the source and target grid/mesh geometries and does not account for
@@ -301,7 +299,10 @@ class Regridder:
         # Note that ESMPy is also able to handle masked data. It is worth investigating
         # how this affects the mathematics and if it can be replicated after the fact
         # using just the weights or if ESMF is doing something we want access to.
-        weight_sums = np.array(self.weight_matrix.sum(axis=1)).flatten()
+        src_inverted_mask = np.where(ma.getmaskarray(src_array), 0, 1)
+        src_mask_matrix = scipy.sparse.diags(src_inverted_mask)
+        masked_weights = self.weight_matrix * src_mask_matrix
+        weight_sums = np.array(masked_weights.sum(axis=1)).flatten()
         # Set the minimum mdtol to be slightly higher than 0 to account for rounding
         # errors.
         mdtol = max(mdtol, 1e-8)
@@ -310,7 +311,7 @@ class Regridder:
         normalisations = np.where(masked_weight_sums == 0, 0, 1 / masked_weight_sums)
         normalisations = ma.array(normalisations, mask=np.logical_not(tgt_mask))
 
-        flat_src = self.src._flatten_array(src_array)
+        flat_src = self.src._flatten_array(ma.getdata(src_array))
         flat_tgt = self.weight_matrix * flat_src
         flat_tgt = flat_tgt * normalisations
         tgt_array = self.tgt._unflatten_array(flat_tgt)
--- a/esmf_regrid/esmf_regridder.py
+++ b/esmf_regrid/esmf_regridder.py
@@ -31,7 +31,14 @@ class GridInfo:
     # TODO: Edit GridInfo so that it is able to handle 2D lat/lon arrays.
 
     def __init__(
-        self, lons, lats, lonbounds, latbounds, crs=None, circular=False, areas=None,
+        self,
+        lons,
+        lats,
+        lonbounds,
+        latbounds,
+        crs=None,
+        circular=False,
+        areas=None,
     ):
         """
         Create a GridInfo object describing the grid.
@@ -120,7 +127,11 @@ class GridInfo:
 
         if circular:
             grid = ESMF.Grid(
-                shape, pole_kind=[1, 1], num_peri_dims=1, periodic_dim=1, pole_dim=0,
+                shape,
+                pole_kind=[1, 1],
+                num_peri_dims=1,
+                periodic_dim=1,
+                pole_dim=0,
             )
         else:
             grid = ESMF.Grid(shape, pole_kind=[1, 1])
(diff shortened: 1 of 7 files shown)


esmf_regrid/tests/unit/esmf_regridder/test_Regridder.py
```diff
@@ -0,0 +1,116 @@
...
+    lon, lat, lon_bounds, lat_bounds = make_grid_args(2, 3)
+    src_grid = GridInfo(lon, lat, lon_bounds, lat_bounds)
+
+    lon, lat, lon_bounds, lat_bounds = make_grid_args(3, 2)
+    tgt_grid = GridInfo(lon, lat, lon_bounds, lat_bounds)
+
+    # Set up the regridder with precomputed weights.
+    rg = Regridder(src_grid, tgt_grid, precomputed_weights=_expected_weights())
+
+    src_array = np.array([[1, 1, 1], [1, 0, 0]])
```
zklaus, line 68: This gives us integer arrays, which raises the question of dtype. Should the source dtype be the result dtype? Intermediate calculations always with double, even if source and target are single precision? Perhaps it's best to avoid those questions for now by using a floating point type here :smile:. As a rule of thumb, I find it useful to always use 1. etc where not explicitly integers are desired.

esmf_regrid/esmf_regridder.py
```diff
@@ -290,27 +290,25 @@ def regrid(self, src_array, mdtol=1):
...
-        weight_sums = np.array(self.weight_matrix.sum(axis=1)).flatten()
+        src_inverted_mask = np.where(ma.getmaskarray(src_array), 0, 1)
+        src_inverted_mask = self.src._flatten_array(src_inverted_mask)
+        weight_sums = self.weight_matrix * src_inverted_mask
         # Set the minimum mdtol to be slightly higher than 0 to account for rounding
         # errors.
         mdtol = max(mdtol, 1e-8)
-        tgt_mask = weight_sums >= 1 - mdtol
+        tgt_mask = weight_sums > 1 - mdtol
         masked_weight_sums = weight_sums * tgt_mask.astype(int)
```
zklaus, line 300: Is it too late to ask why we cast this to `int`?

esmf_regrid/esmf_regridder.py
```diff
@@ -290,27 +290,25 @@ def regrid(self, src_array, mdtol=1):
...
-        normalisations = np.where(masked_weight_sums == 0, 0, 1 / masked_weight_sums)
+        if norm_type == "FRACAREA":
+            normalisations = np.where(
+                masked_weight_sums == 0, 0, 1 / masked_weight_sums
```
zklaus, line 303: This gives a division by zero error in the tests. Not sure why, but would definitely suggest to make the dtype clear:
```suggestion
                masked_weight_sums == 0., 0., 1. / masked_weight_sums
```

esmf_regrid/esmf_regridder.py
```diff
@@ -290,27 +290,25 @@ def regrid(self, src_array, mdtol=1):
...
-        # between the source and target grid/mesh geometries and does not account for
-        # masked data, though it ought to be possible to extend the functionality to
-        # handle masked data.
-        #
-        # Note that ESMPy is also able to handle masked data. It is worth investigating
-        # how this affects the mathematics and if it can be replicated after the fact
-        # using just the weights or if ESMF is doing something we want access to.
-        weight_sums = np.array(self.weight_matrix.sum(axis=1)).flatten()
+        src_inverted_mask = np.where(ma.getmaskarray(src_array), 0, 1)
+        src_inverted_mask = self.src._flatten_array(src_inverted_mask)
```
zklaus, line 294:
```suggestion
        src_inverted_mask = self.src._flatten_array(~ma.getmaskarray(src_array))
```

zklaus (changes requested):

esmf_regrid/esmf_regridder.py
stephenworsley: I think the need for casting this was due to tgt_mask being bool type which didn't multiply well. Casting it as int may well be slightly inappropriate, my intuition was that it would cause less ambiguity with levels of precision, but I'm not really sure what the consequences of casting it to int vs float would be.

zklaus: I think the multiplication will cast the mask to the type of `weight_sums` right away. In principle, multiplication with a bool array should be fine.

esmf_regrid/esmf_regridder.py
stephenworsley: I've decided to change the logic here slightly. The end result is that it replaces all instances of 0 with 1 before division. This shouldn't make a difference because these values will be divided by 0 anyway (since they represent cells which are not mapped to by any other cell).

esmf_regrid/tests/integration/esmf_regridder/test_Regridder.py
```diff
@@ -0,0 +1,67 @@
...
+    src_data = np.array(
+        [
+            [1.0, 1.0, 1.0],
+            [1.0, 0.0, 0.0],
+        ],
+    )
+    src_mask = np.array(
+        [
+            [1, 0, 0],
+            [0, 0, 0],
```
zklaus, line 26:
```suggestion
            [True, False, False],
            [False, False, False],
```

zklaus: I find this a bit clearer. Also, it should give us the right dtype.

esmf_regrid/esmf_regridder.py
```diff
@@ -266,51 +266,52 @@ def __init__(self, src, tgt, precomputed_weights=None):
...
-        masked_weight_sums = weight_sums * tgt_mask.astype(int)
-        normalisations = np.where(masked_weight_sums == 0, 0, 1 / masked_weight_sums)
+        tgt_mask = weight_sums > 1 - mdtol
+        masked_weight_sums = weight_sums * tgt_mask
+        if norm_type == "fracarea":
+            normalisations = 1 / np.where(
+                masked_weight_sums == 0.0, 1.0, masked_weight_sums
+            )
+        elif norm_type == "dstarea":
+            normalisations = np.ones(self.tgt.size())
```
zklaus, line 309:
```suggestion
        normalisations = np.ones(self.tgt.size())
        if norm_type == "fracarea":
            normalisations[tgt_mask] /= masked_weight_sums[tgt_mask]
        elif norm_type == "dstarea":
            pass 
```

esmf_regrid/esmf_regridder.py
```diff
@@ -266,51 +266,52 @@ def __init__(self, src, tgt, precomputed_weights=None):
...
+        else:
+            raise ValueError(f'Normalisation type "{norm_type}" is not supported')
```
zklaus, line 311: Maybe add a test for invalid norm_type?

zklaus (approved):

PR #25, opened 2020-12-17
Iris scheme

zklaus:
Add iris scheme for easy access to the regridder from iris.


esmf_regrid/_grid.py
```diff
@@ -0,0 +1,179 @@
...
+import ESMF
+from iris.exceptions import CellMeasureNotFoundError
+import numpy as np
+
+
+def _bounds_cf_to_simple_1d(cf_bounds):
+    assert (cf_bounds[1:, 0] == cf_bounds[:-1, 1]).all()
+    simple_bounds = np.empty((cf_bounds.shape[0]+1,), dtype=np.float64)
+    simple_bounds[:-1] = cf_bounds[:, 0]
+    simple_bounds[-1] = cf_bounds[-1, 1]
```
stephenworsley, line 13: Compare the iris.coords.Coord methods [is_contiguous](https://github.com/SciTools/iris/blob/077df3ec15530255df80c2fc330eb21998e24769/lib/iris/coords.py#L1732) and [contiguous_bounds](https://github.com/SciTools/iris/blob/077df3ec15530255df80c2fc330eb21998e24769/lib/iris/coords.py#L1762). Notably, the Iris version allows for a degree of tolerance. It might be worth incorporating the Iris versions for consistencies sake.

esmf_regrid/schemes.py
```diff
@@ -1,19 +1,34 @@
...
+from .esmf_regridder import Regridder
+
 __all__ = [
-    "ESMFAreaWeighted",
+    "ESMF",
 ]
 
 
-class ESMFAreaWeighted:
+class ESMF:
```
stephenworsley, line 14: This may cause confusion since there is already a package being imported named ESMF.

stephenworsley:
For reference, on the subject of coppying over coordinates and metadata to the new cube, in iris this seems to be done by something like the [following method](https://github.com/SciTools/iris/blob/48b8e2488d4576b1afacefafffad3a609ec903d7/lib/iris/analysis/_regrid.py#L819). This method is tailored to the case of regridding  rectilinear grids to rectilinear grids (which should be sufficient for the current use cases). Notably, it has been made a static function so that it can be reused in other regridders (for example [here](https://github.com/SciTools/iris/blob/48b8e2488d4576b1afacefafffad3a609ec903d7/lib/iris/experimental/regrid.py#L949)). Also of note is that this code has been mostly copied and slightly tweaked [here](https://github.com/SciTools/iris/blob/48b8e2488d4576b1afacefafffad3a609ec903d7/lib/iris/experimental/regrid.py#L1225) for the case of regridding one dimensional unstructured data to rectilinear grids.

My thinking is that we will probably want to eventually write a more general version of this method which handles any combination of rectilinear/curvilinear/unstructured grids, though it may be possible to use this method directly innitialy while we are restricted to rectilinear grids. The one thing which might make this tricky is the regrid_callback argument, which is used to apply the same regridding to certain aux coords. This may be tricky to adapt to our case because the arguments our regridder takes will be slightly different than maybe this code is expecting. It may be worthwhile ignoring these cases for the time being since they seem somewhat of an edge case, perhaps with something like the following code:
```python
def null_callback(*ags, **kwargs):
    raise iris.exceptions.NotYetImplementedError

slim_rect_create_cube = functools.partial(
    RectilinearRegridder._create_cube,
    sample_grid_x=None,
    sample_grid_y=None,
    regrid_callback=null_callback,
)
```

Alternatively, there may also be a case for starting with a more minimal approach to coord copying like that in the curvilinear to rectilinear regridder which omits much of the _create_cube functionality. The code for this can be seen [here](https://github.com/SciTools/iris/blob/48b8e2488d4576b1afacefafffad3a609ec903d7/lib/iris/analysis/_regrid.py#L356-L363).

PR #26, opened 2020-12-17
Iris scheme sparse

zklaus:
Iris scheme using sparse


zklaus:
The following is @stephenworsley's comment copied over from https://github.com/SciTools-incubator/iris-esmf-regrid/pull/25#issuecomment-781240259

For reference, on the subject of coppying over coordinates and metadata to the new cube, in iris this seems to be done by something like the [following method](https://github.com/SciTools/iris/blob/48b8e2488d4576b1afacefafffad3a609ec903d7/lib/iris/analysis/_regrid.py#L819). This method is tailored to the case of regridding  rectilinear grids to rectilinear grids (which should be sufficient for the current use cases). Notably, it has been made a static function so that it can be reused in other regridders (for example [here](https://github.com/SciTools/iris/blob/48b8e2488d4576b1afacefafffad3a609ec903d7/lib/iris/experimental/regrid.py#L949)). Also of note is that this code has been mostly copied and slightly tweaked [here](https://github.com/SciTools/iris/blob/48b8e2488d4576b1afacefafffad3a609ec903d7/lib/iris/experimental/regrid.py#L1225) for the case of regridding one dimensional unstructured data to rectilinear grids.

My thinking is that we will probably want to eventually write a more general version of this method which handles any combination of rectilinear/curvilinear/unstructured grids, though it may be possible to use this method directly innitialy while we are restricted to rectilinear grids. The one thing which might make this tricky is the regrid_callback argument, which is used to apply the same regridding to certain aux coords (specifically, ones which have derived coords associated with them). This may be tricky to adapt to our case because the arguments our regridder takes will be slightly different than maybe this code is expecting. It may be worthwhile ignoring these cases for the time being since they seem somewhat of an edge case, perhaps with something like the following code:
```python
def null_callback(*ags, **kwargs):
    raise iris.exceptions.NotYetImplementedError

slim_rect_create_cube = functools.partial(
    RectilinearRegridder._create_cube,
    sample_grid_x=None,
    sample_grid_y=None,
    regrid_callback=null_callback,
)
```

Alternatively, there may also be a case for starting with a more minimal approach to coord copying like that in the curvilinear to rectilinear regridder which omits much of the _create_cube functionality. The code for this can be seen [here](https://github.com/SciTools/iris/blob/48b8e2488d4576b1afacefafffad3a609ec903d7/lib/iris/analysis/_regrid.py#L356-L363).

PR #28, opened 2021-02-12
Extra dimension handling (previously Dimension priority reorder)

### LINKED ISSUE #27: Lat-lon dimension ordering switched
## 🐛 Bug Report

As demonstrated by the code included at the end of #24, data on grids is returned with the dimensions ordered longitude-latitude. According to CF spec, the tendancy is for dimensions to be ordered latitude-longitude. This is the convention used by Cartopy for example, which is why the transpose needed to be taken in order to plot properly with Cartopy. The code should be changed to respect the default CF dimension ordering.

stephenworsley:
### Original Description

Fixes #27. This PR changes the order in which the regridder expects the dimensions of incoming and outgoing data to be ordered to better reflect those described by [CF conventions](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#dimensions). That is, data is expected to be ordered such that longitude is the last dimension and latitude is the second to last dimension. A side effect of the more comprehensive approach I have taken is that it is now possible to regrid data with additional non-horizontal dimensions (which has been accounted for with additional tests).

This reordering is largely confined to the code in the `_flatten_array`/`_unflatten_array` methods. These have become more complex than before, which I believe is due to a mismatch in the way ESMF orders dimensions (this is possibly due to ESMF written in Fortran). With that said, the code for `_flatten_array`/`_unflatten_array` in `MeshInfo` and `GridInfo` is now unified.

It's worth noting, and is demonstrated by initial commits, that the all tests still work so long as the original data is transposed.

### New Description

The part of this PR concerned with the reordering of latitude/longitude has now been handled by #30. This PR has been rebased on top of that and now this PR is only concerned with the handling of additional data dimensions.

This PR allows for the regridding of data with dimensions not associated with the horizontal grid/mesh. These extra dimensions must precede the grid/mesh dimensions (as suggested by [CF conventions](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#dimensions)). We effectively broadcast the regridding operation defined on the grid/mesh dimensions over all extra dimensions.

### DIFF
--- a/esmf_regrid/esmf_regridder.py
+++ b/esmf_regrid/esmf_regridder.py
@@ -304,7 +304,7 @@ class Regridder:
                 f"got an array with shape ending in {main_shape}."
             )
         extra_shape = array_shape[: -self.src.dims]
-        extra_size = max(1, sum(extra_shape))
+        extra_size = max(1, np.prod(extra_shape))
         src_inverted_mask = self.src._flatten_array(~ma.getmaskarray(src_array))
         weight_sums = self.weight_matrix * src_inverted_mask
         # Set the minimum mdtol to be slightly higher than 0 to account for rounding
--- a/esmf_regrid/esmf_regridder.py
+++ b/esmf_regrid/esmf_regridder.py
@@ -175,10 +175,10 @@ class GridInfo:
         return 1
 
     def _flatten_array(self, array):
-        return array.flatten(order="F")
+        return array.reshape(-1, (self.size()), order="F").T
 
-    def _unflatten_array(self, array):
-        return array.reshape(self.shape, order="F")
+    def _unflatten_array(self, array, extra_dims):
+        return array.T.reshape(extra_dims + self.shape, order="F")
 
 
 def _get_regrid_weights_dict(src_field, tgt_field):
@@ -303,6 +303,8 @@ class Regridder:
                 f"Expected an array whose shape ends in {self.src.shape}, "
                 f"got an array with shape ending in {main_shape}."
             )
+        extra_shape = array_shape[:-self.src.dims]
+        extra_size = max(1, sum(extra_shape))
         src_inverted_mask = self.src._flatten_array(~ma.getmaskarray(src_array))
         weight_sums = self.weight_matrix * src_inverted_mask
         # Set the minimum mdtol to be slightly higher than 0 to account for rounding
@@ -310,7 +312,7 @@ class Regridder:
         mdtol = max(mdtol, 1e-8)
         tgt_mask = weight_sums > 1 - mdtol
         masked_weight_sums = weight_sums * tgt_mask
-        normalisations = np.ones(self.tgt.size())
+        normalisations = np.ones([self.tgt.size(), extra_size])
         if norm_type == "fracarea":
             normalisations[tgt_mask] /= masked_weight_sums[tgt_mask]
         elif norm_type == "dstarea":
@@ -322,5 +324,5 @@ class Regridder:
         flat_src = self.src._flatten_array(ma.filled(src_array, 0.0))
         flat_tgt = self.weight_matrix * flat_src
         flat_tgt = flat_tgt * normalisations
-        tgt_array = self.tgt._unflatten_array(flat_tgt)
+        tgt_array = self.tgt._unflatten_array(flat_tgt, extra_shape)
         return tgt_array
--- a/esmf_regrid/esmf_regridder.py
+++ b/esmf_regrid/esmf_regridder.py
@@ -174,10 +174,25 @@ class GridInfo:
     def _index_offset(self):
         return 1
 
-    def _flatten_array(self, array):
+    def _array_to_matrix(self, array):
+        """
+        Reshape data to a form that is compatible with weight matrices.
+
+        The data should be presented in the form of a matrix (i.e. 2D) in order
+        to be compatible with the weight matrix.
+        Weight matrices deriving from ESMF use fortran ordering when flattening
+        grids to determine cell indices so we use the same order for reshaping.
+        We then take the transpose so that matrix multiplication happens over
+        the appropriate axes.
+        """
         return array.reshape(-1, (self.size()), order="F").T
 
-    def _unflatten_array(self, array, extra_dims):
+    def _matrix_to_array(self, array, extra_dims):
+        """
+        Reshape data to restore original dimensions.
+
+        This is the inverse operation of `_array_to_matrix`.
+        """
         return array.T.reshape(extra_dims + self.shape, order="F")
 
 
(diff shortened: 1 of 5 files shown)


stephenworsley:
When running the code described in https://github.com/SciTools-incubator/iris-esmf-regrid/pull/24#issuecomment-767687876 and changing the lines:

```python
out_cube.add_dim_coord(lon_coord, 0)
out_cube.add_dim_coord(lat_coord, 1)
```
to
```python
out_cube.add_dim_coord(lon_coord, 1)
out_cube.add_dim_coord(lat_coord, 0)
```
the plots turn out the same as in the linked comment.

Alternately, when plotting with the commented out cartopy code, replacing the line:
```python
plt.pcolormesh(grid_lons, grid_lats, out_data.T)
```
with

```python
plt.pcolormesh(grid_lons, grid_lats, out_data)
```
also yields the equivalent plots.

esmf_regrid/esmf_regridder.py
```diff
@@ -173,10 +175,10 @@ def _index_offset(self):
         return 1
 
     def _flatten_array(self, array):
-        return array.flatten()
+        return array.reshape(-1, (self.size()), order="F").T
```
zklaus, line 178: Perhaps `self.size` should be a `@property`? In any case, I think the `()` around `self.size()` don't do anything here, right?

I suggest to remove the superfluous `()` and consider making size a property, though perhaps in a different PR.

zklaus: Two more questions:

1. The transposition is here to allow for the application to be executed by matrix-vector multiplication *from the left* instead of from the right, correct?
2. Do we really need to specify the order and are we sure this does what we want? I was trying to read a bit more about `reshape`'s handling of the order parameter in [the documentation](https://numpy.org/doc/stable/reference/generated/numpy.reshape.html), but I can't say that it fills me with confidence. Note for example:
>  Note there is no guarantee of the memory layout (C- or Fortran- contiguous) of the returned array.

And have a look at the explanation of the order parameter in the Notes section with the examples.

If you are sure this is all right, I suggest adding a note either as a comment or in a docstring that explains why this is here and how it helps us.

esmf_regrid/esmf_regridder.py
```diff
@@ -294,14 +296,23 @@ def regrid(self, src_array, norm_type="fracarea", mdtol=1):
             A numpy array whose shape is compatible with self.tgt.
 
         """
+        array_shape = src_array.shape
+        main_shape = array_shape[-self.src.dims :]
+        if main_shape != self.src.shape:
+            raise ValueError(
+                f"Expected an array with whose shape ends in {self.src.shape}, "
```
zklaus, line 303: Typo: Remove `with`

esmf_regrid/experimental/unstructured_regrid.py
```diff
@@ -116,13 +118,13 @@ def make_esmf_field(self):
...
         return self.esi
 
     def _flatten_array(self, array):
-        return array
+        return array.reshape(-1, (self.size()), order="F").T
 
-    def _unflatten_array(self, array):
-        return array
+    def _unflatten_array(self, array, extra_dims):
+        return array.T.reshape(extra_dims + self.shape, order="F")
```
zklaus, line 130: `_flatten_array` and `_unflatten_array` are identical to the versions above, so the same comments apply. Moreover, perhaps this is reason enough to have a common baseclass? `size` can be made identical, I believe, with `prod(self.shape)` as well.

zklaus (commented):
Only minor comments. Nice job.

zklaus:
Just saw that this is split up, partly in #30. Maybe we will close this PR?

stephenworsley:
> 
> 
> Just saw that this is split up, partly in #30. Maybe we will close this PR?

I've rebased to include #30, there may still be a case for having a fresh PR for the sake of GitHub clarity, but it should at least be mergable in this state.

esmf_regrid/esmf_regridder.py
jamesp: @stephenworsley let's talk this through today, I certainly agree with @zklaus that a comment here explaining why it's necessary.  In general, anywhere I put a `.T` is often a good signal that a comment is needed... I few weeks later often forget why it was the right thing to do !

jamesp: Great additional doc @stephenworsley 👍

jamesp (approved):

jamesp:
Hey @zklaus, let us know what you think of the changes and additional tests - LGTM if you're happy! 👍

esmf_regrid/esmf_regridder.py
zklaus: The `()` can still go away, right?

esmf_regrid/experimental/unstructured_regrid.py
```diff
@@ -123,8 +123,23 @@ def size(self):
...
+        Reshape data to a form that is compatible with weight matrices.
+
+        The data should be presented in the form of a matrix (i.e. 2D) in order
+        to be compatible with the weight matrix.
+        Weight matrices deriving from ESMF use fortran ordering when flattening
+        grids to determine cell indices so we use the same order for reshaping.
+        We then take the transpose so that matrix multiplication happens over
+        the appropriate axes.
+        """
+        return array.reshape(-1, (self.size()), order="F").T
```
zklaus, line 137: Remove superfluous `()`.

zklaus (changes requested):
Super minor comment, but I'd really like to see those `()` go because they always trick me into thinking there is a tuple there, which is confusing.

esmf_regrid/esmf_regridder.py
```diff
@@ -185,7 +186,7 @@ def _array_to_matrix(self, array):
         We then take the transpose so that matrix multiplication happens over
         the appropriate axes.
         """
-        return array.reshape(-1, (self.size()), order="F").T
+        return array.reshape(-1, (self.size), order="F").T
```
zklaus, line 189: I like the property :smile: but I was talking about the parentheses that are now left around `self.size`, ie
```suggestion
        return array.reshape(-1, self.size, order="F").T
```

esmf_regrid/experimental/unstructured_regrid.py
```diff
@@ -134,7 +135,7 @@ def _array_to_matrix(self, array):
         We then take the transpose so that matrix multiplication happens over
         the appropriate axes.
         """
-        return array.reshape(-1, (self.size()), order="F").T
+        return array.reshape(-1, (self.size), order="F").T
```
zklaus, line 138: ditto
```suggestion
        return array.reshape(-1, self.size, order="F").T
```

zklaus (changes requested):

zklaus (approved):

PR #29, opened 2021-02-21
fix bash octal date

bjlittle:
This PR is a fix to the `.cirrus.yml`, where the output of `$(date +%U)` returns `week number of year, with Sunday as first day of week (00..53)` can be interpreted as an octal number that is out of range e.g., `08`.

This causes `bash` to raise and error, and subsequently for all `cirrus-ci` tasks to fail.


stephenworsley (approved):
Looks good!

PR #30, opened 2021-02-24
Latlon reorder simplified

stephenworsley:
Split from #28 to reduce the complexity of the individual PRs. This does not handle any extra dimensions and is only responsible for reordering the existing dimensions and adding a check.

### DIFF
--- a/esmf_regrid/esmf_regridder.py
+++ b/esmf_regrid/esmf_regridder.py
@@ -175,10 +175,10 @@ class GridInfo:
         return 1
 
     def _flatten_array(self, array):
-        return array.reshape(-1, (self.size()), order="F").T
+        return array.flatten(order="F")
 
-    def _unflatten_array(self, array, extra_dims):
-        return array.T.reshape(extra_dims + self.shape, order="F")
+    def _unflatten_array(self, array):
+        return array.reshape(self.shape, order="F")
 
 
 def _get_regrid_weights_dict(src_field, tgt_field):
@@ -303,8 +303,6 @@ class Regridder:
                 f"Expected an array with whose shape ends in {self.src.shape}, "
                 f"got an array with shape ending in {main_shape}."
             )
-        extra_shape = array_shape[: -self.src.dims]
-        extra_size = max(1, sum(extra_shape))
         src_inverted_mask = self.src._flatten_array(~ma.getmaskarray(src_array))
         weight_sums = self.weight_matrix * src_inverted_mask
         # Set the minimum mdtol to be slightly higher than 0 to account for rounding
@@ -312,7 +310,7 @@ class Regridder:
         mdtol = max(mdtol, 1e-8)
         tgt_mask = weight_sums > 1 - mdtol
         masked_weight_sums = weight_sums * tgt_mask
-        normalisations = np.ones([self.tgt.size(), extra_size])
+        normalisations = np.ones(self.tgt.size())
         if norm_type == "fracarea":
             normalisations[tgt_mask] /= masked_weight_sums[tgt_mask]
         elif norm_type == "dstarea":
@@ -324,5 +322,5 @@ class Regridder:
         flat_src = self.src._flatten_array(ma.filled(src_array, 0.0))
         flat_tgt = self.weight_matrix * flat_src
         flat_tgt = flat_tgt * normalisations
-        tgt_array = self.tgt._unflatten_array(flat_tgt, extra_shape)
+        tgt_array = self.tgt._unflatten_array(flat_tgt)
         return tgt_array
--- a/esmf_regrid/esmf_regridder.py
+++ b/esmf_regrid/esmf_regridder.py
@@ -78,9 +78,11 @@ class GridInfo:
             self.crs = crs
         self.circular = circular
         self.areas = areas
+        self.shape = (len(lats), len(lons))
+        self.dims = 2
 
     def _as_esmf_info(self):
-        shape = np.array([len(self.lats), len(self.lons)])
+        shape = np.array(self.shape)
 
         if self.circular:
             adjustedlonbounds = self.lonbounds[:-1]
@@ -173,10 +175,10 @@ class GridInfo:
         return 1
 
     def _flatten_array(self, array):
-        return array.flatten()
+        return array.reshape(-1, (self.size()), order="F").T
 
-    def _unflatten_array(self, array):
-        return array.reshape((len(self.lons), len(self.lats)))
+    def _unflatten_array(self, array, extra_dims):
+        return array.T.reshape(extra_dims + self.shape, order="F")
 
 
 def _get_regrid_weights_dict(src_field, tgt_field):
@@ -294,6 +296,9 @@ class Regridder:
             A numpy array whose shape is compatible with self.tgt.
 
         """
+        array_shape = src_array.shape
+        extra_shape = array_shape[:-self.src.dims]
+        extra_size = max(1, sum(extra_shape))
         src_inverted_mask = self.src._flatten_array(~ma.getmaskarray(src_array))
         weight_sums = self.weight_matrix * src_inverted_mask
         # Set the minimum mdtol to be slightly higher than 0 to account for rounding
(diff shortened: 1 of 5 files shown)


bjlittle:
@stephenworsley Simply rebase your branch with the latest changes from `main` to pick up the fixes for `cirrus-ci` :+1:

esmf_regrid/esmf_regridder.py
```diff
@@ -294,6 +296,13 @@ def regrid(self, src_array, norm_type="fracarea", mdtol=1):
             A numpy array whose shape is compatible with self.tgt.
 
         """
+        array_shape = src_array.shape
+        main_shape = array_shape[-self.src.dims :]
+        if main_shape != self.src.shape:
```
stephenworsley, line 301: It's worth considering the case where one (or more) of the grid dimensions is a scalar. We may end up comparing (n) with (1, n) and raising an error where we would otherwise want to be lenient.

esmf_regrid/tests/unit/esmf_regridder/test_Regridder.py
```diff
@@ -66,16 +66,15 @@ def test_Regridder_regrid():
     # Set up the regridder with precomputed weights.
```
abooton, line 66: In order to understand, and hence check that we have matched up the esmf computed weights to the correct location in our Python array, I have created a slightly simpler (3x2) array test, that can be worked through manually and checked on paper!   In the following example.  Having achieved this I am confident that we have the somewhat intricate conversions from esmf weight location to Iris location correct (referenced by 'F' ordering, but unravelled in a 'C' ordering, with elements reference per target-src point ordering)!

Given that this conversion is non-trivial to think through understnad and correct, I was wondering if we could include this simpler example alongside the more complex one already implemented.  Note that it is not aiming to check the entire initialisation method, but more explicitly obtaining the weights and converting them into use with the following code.

I think it would have been ideal to re-order the weights into our "Iris format" idexes at this point, rather than flattening the data in the Fortan ordering during the regridding.  However, I have confirmed that the results are correct offline.  (I'll add a note about this in another comment).  Therefore, if we could add this extra test here, I think the code is good to merge, and further improvements could be made once the correct test results are committed.

def test_Regridder_init_small():
    lon_bounds = np.array([0, 10, 20])
    lat_bounds = np.array([0, 10, 20, 30])
    lon, lat = _get_points(lon_bounds), _get_points(lat_bounds)
    src_grid = GridInfo(lon, lat, lon_bounds, lat_bounds)
    assert (src_grid.shape == (3,2))
    assert (src_grid._index_offset() == 1)

    lon_bounds = np.array([0, 20])
    lat_bounds = np.array([0, 10, 30])
    lon, lat = _get_points(lon_bounds), _get_points(lat_bounds)
    tgt_grid = GridInfo(lon, lat, lon_bounds, lat_bounds)
    assert (tgt_grid.shape == (2, 1,))
    assert (tgt_grid._index_offset() == 1)

    rg = Regridder(src_grid, tgt_grid)

    result = rg.weight_matrix
    
    weights_dict = {}
    weights_dict['row_dst'] = np.array([1, 1, 1, 1, 2, 2, 2, 2], dtype=np.int32) - src_grid._index_offset()
    weights_dict['col_src'] = np.array([1, 2, 4, 5, 2, 3, 5, 6], dtype=np.int32) - tgt_grid._index_offset()
    weights_dict['weights'] = np.array([0.4962897, 0.0037103, 0.4962897, 0.0037103 , 0.25484249, 0.24076863, 0.25484249, 0.24076863])

    expected_weights = scipy.sparse.csr_matrix((weights_dict['weights'], (weights_dict['row_dst'], weights_dict['col_src']))) #, shape=shape)
    assert np.allclose(result.toarray(), expected_weights.toarray())

esmf_regrid/tests/unit/esmf_regridder/test_Regridder.py
```diff
@@ -86,35 +85,111 @@ def test_Regridder_regrid():
...
+    lon, lat = _get_points(lon_bounds), _get_points(lat_bounds)
+    src_grid = GridInfo(lon, lat, lon_bounds, lat_bounds)
+    assert src_grid.shape == (3, 2)
+    assert src_grid._index_offset() == 1
+
+    lon_bounds = np.array([0, 20])
+    lat_bounds = np.array([0, 10, 30])
+    lon, lat = _get_points(lon_bounds), _get_points(lat_bounds)
+    tgt_grid = GridInfo(lon, lat, lon_bounds, lat_bounds)
+    assert tgt_grid.shape == (
```
abooton, line 162: Just curious as to why you wrote the tgt shape out like this?

esmf_regrid/tests/unit/esmf_regridder/test_Regridder.py
```diff
@@ -86,35 +85,111 @@ def test_Regridder_regrid():
...
+
+def test_Regridder_init_small():
+    """
+    Simplified test for :meth:`~esmf_regrid.esmf_regridder.Regridder.regrid`.
+
+    This test is designed to be simple enough to generate predictable weights.
+    With predictable weights it is easier to check that the weights are
+    associated with the correct indices.
+    """
+    # The following ASCII visualisation describes the source and target grids
```
abooton, line 134: I like your explanation :)

esmf_regrid/tests/unit/esmf_regridder/test_Regridder.py
```diff
@@ -86,35 +85,111 @@ def test_Regridder_regrid():
...
+    # Analysis of the weights dict returned by ESMF confirms that this is
+    # indeed the indexing which ESMF assigns these grids.
+    #
+    # 30  +---+---+       +-------+
+    #     | 3 | 6 |       |       |
+    # 20  +---+---+       |   2   |
+    #     | 2 | 5 |       |       |
+    # 10  +---+---+       +-------+
+    #     | 1 | 4 |       |   1   |
+    #  0  +---+---+       +-------+
```
abooton, line 145: The tgt cell 1, also overlaps with a tiny area of src cells 2 and 5. (this confused me for a while)

zklaus (commented):
Minor comments. Nice job.

abooton:
### Additional checking to ensure the main regridding method is giving the correct results
Confirming that the weights are being applied to the correct data value is particularly tricky due to several factors:
1) esmf indexing has "F" column major-indexing, but Iris uses pythons "C" ordering
2) The order of the weights in the array is flattened src-cell overlap per target cell, so is not a simple row-column order of cells in an array
3) Care has to be taken to consider indexing in rows and columns, because in Iris, "x" is usually referencing the x-coordinate which corresponds to the columns.  Hence it is easy to confuse Iris ordering and src-tgt configurations.

Given the above I found confirming that the correct weights were being applied to the correct data values particularly challenging, especially as I felt it may be possible to accidentally miscalling the resulting field incorrectly, despite the result looking pretty good.

Therefore, I have tested the regridding method in the following way:

1) Refactored the existing (core) Iris area-weighted regridding code such that the weights are calculated in the prepare step, and and the regridding is performed in a single step via whole-array (numpy array) manipulations.  (Thus the _prepare and_perform steps are akin to those used in iris-esmf methods)
See https://github.com/SciTools/iris/compare/master...abooton:alter_awr_to_take_w_i_arrays

2) I rejigged the weights to be in the format that esmf outputs.  And replaced the _perform regridding method with that used in here
See  https://github.com/abooton/iris/pull/new/check_esmf_awr_method
Enough of the tests passed for me to be convinced that the method as coded does give the correct answers.  Those that were failing were  accommodating the extra dimensions, scalar inputs and missing values - all of which are due to be addressed at a later date.

esmf_regrid/esmf_regridder.py
abooton: Having discussed this, we think this is needed but will be done in a separate PR as it doesn't impact the answers here.

esmf_regrid/tests/unit/esmf_regridder/test_Regridder.py
stephenworsley: Added with additional documentation.

esmf_regrid/tests/unit/esmf_regridder/test_Regridder.py
stephenworsley: This is addressed in a comment below.

esmf_regrid/esmf_regridder.py
```diff
@@ -294,6 +296,13 @@ def regrid(self, src_array, norm_type="fracarea", mdtol=1):
...
+            raise ValueError(
+                f"Expected an array with whose shape ends in {self.src.shape}, "
```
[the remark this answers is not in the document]
abooton: Done

abooton (approved):
This is looking good to me.  Just to note that the @property work is to be addressed in another PR, as discussed in the comments above.

PR #32, opened 2021-03-01
Unstructured Scheme - Basic GridInfo Handling

stephenworsley:
This copies the most basic functionality from #26 to integrate rectilinear grid handling. This allows development of the unstructured regridder to continue.

### DIFF
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -2,13 +2,21 @@
 
 import iris
 from iris.analysis._interpolation import get_xy_coords
+import numpy as np
 
 # from numpy import ma
 
-from esmf_regrid.esmf_regridder import Regridder
+from esmf_regrid.esmf_regridder import GridInfo, Regridder
 
-# from esmf_regrid.esmf_regridder import GridInfo
 # from esmf_regrid.experimental.unstructured_regrid import MeshInfo
+
+# Taken from PR #26
+def _bounds_cf_to_simple_1d(cf_bounds):
+    assert (cf_bounds[1:, 0] == cf_bounds[:-1, 1]).all()
+    simple_bounds = np.empty((cf_bounds.shape[0] + 1,), dtype=np.float64)
+    simple_bounds[:-1] = cf_bounds[:, 0]
+    simple_bounds[-1] = cf_bounds[-1, 1]
+    return simple_bounds
 
 
 def _get_mesh_and_dim(cube):
@@ -28,7 +36,17 @@ def _cube_to_MeshInfo(cube):
 def _cube_to_GridInfo(cube):
     # Returns a GridInfo object describing the horizontal grid of the cube.
     # This may be inherited from code written for the rectilinear regridding scheme.
-    pass
+    lat = cube.coord("latitude")
+    lon = cube.coord("longitude")
+    # TODO: accomodate other x/y coords.
+    # TODO: perform checks on lat/lon.
+    return GridInfo(
+        lon.points,
+        lat.points,
+        _bounds_cf_to_simple_1d(lon.bounds),
+        _bounds_cf_to_simple_1d(lat.bounds),
+        circular=lon.circular,
+    )
 
 
 # def _regrid_along_dims(regridder, data, src_dim, mdtol):
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -35,12 +35,16 @@ def _cube_to_MeshInfo(cube):
 
 
 def _cube_to_GridInfo(cube):
+    # This is a simplified version of an equivalent function/method in PR #26.
+    # It is anticipated that this function will be replaced by the one in PR #26.
+    #
     # Returns a GridInfo object describing the horizontal grid of the cube.
     # This may be inherited from code written for the rectilinear regridding scheme.
     lat = cube.coord("latitude")
     lon = cube.coord("longitude")
     # TODO: accomodate other x/y coords.
     # TODO: perform checks on lat/lon.
+    #  Checks may cover units, coord systems (e.g. rotated pole), contiguous bounds.
     return GridInfo(
         lon.points,
         lat.points,
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -42,6 +42,9 @@ def _cube_to_GridInfo(cube):
     # This may be inherited from code written for the rectilinear regridding scheme.
     lat = cube.coord("latitude")
     lon = cube.coord("longitude")
+    # Ensure coords come from a proper grid.
+    assert isinstance(lat, iris.coords.DimCoord)
+    assert isinstance(lon, iris.coords.DimCoord)
     # TODO: accomodate other x/y coords.
     # TODO: perform checks on lat/lon.
     #  Checks may cover units, coord systems (e.g. rotated pole), contiguous bounds.
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -10,6 +10,7 @@ from esmf_regrid.esmf_regridder import GridInfo, Regridder
 
 # from esmf_regrid.experimental.unstructured_regrid import MeshInfo
 
+
 # Taken from PR #26
 def _bounds_cf_to_simple_1d(cf_bounds):
     assert (cf_bounds[1:, 0] == cf_bounds[:-1, 1]).all()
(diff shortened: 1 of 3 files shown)


esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -28,7 +37,17 @@ def _cube_to_MeshInfo(cube):
 def _cube_to_GridInfo(cube):
     # Returns a GridInfo object describing the horizontal grid of the cube.
     # This may be inherited from code written for the rectilinear regridding scheme.
```
stephenworsley, line 39: We should note that we anticipate this function mirroring an equivalent function (likely with a different name) in #26.

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_cube_to_GridInfo.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def test_cube_to_GridInfo():
+    """Basic test for :func:`~esmf_regrid.experimental.unstructured_scheme._cube_to_GridInfo`."""
+    n_lats = 5
+    n_lons = 6
+
+    cube = _grid_cube_global(n_lats, n_lons)
+    gridinfo = _cube_to_GridInfo(cube)
+    _ = gridinfo.make_esmf_field()
```
stephenworsley, line 63: Add a note to explain that thif is to test one of the main functions of GridInfo works without error.

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_cube_to_GridInfo.py
```diff
@@ -0,0 +1,77 @@
...
+from iris.cube import Cube
+import numpy as np
+import scipy.sparse
+
+from esmf_regrid.esmf_regridder import Regridder
+from esmf_regrid.experimental.unstructured_scheme import _cube_to_GridInfo
+
+
+def _grid_cube_global(n_lats, n_lons):
+    lat_span = np.linspace(-90, 90, n_lats * 2 + 1)
```
abooton, line 13: I'm sure you've mentioned this before, but please could you just remind me here (not in the code) what the benefit of using linspace is for calculating the grid.

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_cube_to_GridInfo.py
```diff
@@ -0,0 +1,77 @@
...
+def test_cube_to_GridInfo():
+    """Basic test for :func:`~esmf_regrid.experimental.unstructured_scheme._cube_to_GridInfo`."""
+    n_lats = 5
+    n_lons = 6
+
+    cube = _grid_cube_global(n_lats, n_lons)
+    gridinfo = _cube_to_GridInfo(cube)
+    _ = gridinfo.make_esmf_field()
+
+    # The following test ensures there are no overlapping cells.
```
stephenworsley, line 65: explain what kind of errors would be caught by the choice of input parameters, i.e. switching lat and lon.

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_cube_to_GridInfo.py
```diff
@@ -0,0 +1,77 @@
...
+    rg = Regridder(gridinfo, gridinfo)
+    expected_weights = scipy.sparse.identity(n_lats * n_lons)
+    assert np.array_equal(expected_weights.todense(), rg.weight_matrix.todense())
+
```
stephenworsley, line 69: Separate these into two tests, consider the naming convention (keep consistent with iris).

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -28,7 +37,17 @@ def _cube_to_MeshInfo(cube):
...
-    pass
+    lat = cube.coord("latitude")
+    lon = cube.coord("longitude")
+    # TODO: accomodate other x/y coords.
+    # TODO: perform checks on lat/lon.
```
abooton, line 43: e.g. checks may include units, rotated poles, points have contiguous bounds

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -28,7 +37,17 @@ def _cube_to_MeshInfo(cube):
 def _cube_to_GridInfo(cube):
     # Returns a GridInfo object describing the horizontal grid of the cube.
     # This may be inherited from code written for the rectilinear regridding scheme.
-    pass
+    lat = cube.coord("latitude")
```
abooton, line 40: Do we need to confirm that the incoming cube is a gridded cube?

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_cube_to_GridInfo.py
abooton, line 65: **Confirming that the test is definitely picking up lats and lons in the reversed order.**
We have just run the tests here by passing in the latitudes and longitudes reversed.  The test failed as expected.
We also passed in the lat lons reversed, but using a square grid that spanned the entire global region. The test fails.

It will pass if we use a square grid of "square" range, e.g. -90 +90 in lats and lons.  
However, this is OK, as our test is coded to fail correctly as described above.

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_cube_to_GridInfo.py
```diff
@@ -0,0 +1,77 @@
...
+    data = np.zeros([n_lats, n_lons])
+    cube = Cube(data)
+    cube.add_dim_coord(lat, 0)
+    cube.add_dim_coord(lon, 1)
+    return cube
+
+
+def test_cube_to_GridInfo():
+    """Basic test for :func:`~esmf_regrid.experimental.unstructured_scheme._cube_to_GridInfo`."""
+    n_lats = 5
```
abooton, line 58: Maybe we should have a test for n_lats=1.  I believe this could be passed in.  Not sure what the final behaviour is!

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_cube_to_GridInfo.py
stephenworsley: doing an equivalent operation with `np.arange` has an issue where it can be sensitive to rounding errors.

esmf_regrid/tests/unit/experimental/unstructured_scheme/test__cube_to_GridInfo.py
```diff
@@ -0,0 +1,108 @@
...
+from iris.coords import DimCoord
+from iris.cube import Cube
+import numpy as np
+import scipy.sparse
+
+from esmf_regrid.esmf_regridder import Regridder
+from esmf_regrid.experimental.unstructured_scheme import _cube_to_GridInfo
+
+
+def _grid_cube_global(n_lats, n_lons):
```
abooton, line 12: This code is very similar to below.  Let's simplify it quickly, and use our usual Iris ordering for the inputs
`def _grid_cube(n_lons, lon_st, lon_end, n_lats, lat_st, lat_end)`

abooton: Much better 👍

esmf_regrid/tests/unit/experimental/unstructured_scheme/test__cube_to_GridInfo.py
```diff
@@ -0,0 +1,94 @@
...
+from iris.cube import Cube
+import numpy as np
+import scipy.sparse
+
+from esmf_regrid.esmf_regridder import Regridder
+from esmf_regrid.experimental.unstructured_scheme import _cube_to_GridInfo
+
+
+def _grid_cube(n_lons, n_lats, lon_bounds, lat_bounds, circular=False):
+    lon_lower, lon_upper = lon_bounds
```
abooton, line 13: Lines 13-17 are identical to 22-27.  And only the `ll_bound_span` needs to be returned in both cases.
Lets refactor and tidy this up - then its looking good.

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -26,9 +35,26 @@ def _cube_to_MeshInfo(cube):
...
+    # It is anticipated that this function will be replaced by the one in PR #26.
+    #
     # Returns a GridInfo object describing the horizontal grid of the cube.
     # This may be inherited from code written for the rectilinear regridding scheme.
-    pass
+    lon = cube.coord("longitude")
+    lat = cube.coord("latitude")
+    # Ensure coords come from a proper grid.
+    assert isinstance(lon, iris.coords.DimCoord)
+    assert isinstance(lat, iris.coords.DimCoord)
```
abooton, line 47: I like this - but I've just had a thought as to whether the scalar passes.  Let's just confirm, and remove these lines if they are more problematic than helpful!

abooton: The scalar test is passing fine with this.
Having discussed this with @stephenworsley , we think this is because the information is added to the `Cube` as a DimCoord, then it becomes a scalar as it has length 1. Hence the test and the data process correctly.

PR #34, opened 2021-03-06
module level setup

bjlittle:
This PR ensures that the `setup` is always executed on `import` and when run as a script.

In addition, the `pre-commit-hooks` version is updated, and additional ignores are added for `flake8`.


stephenworsley (approved):
Looks good.

PR #33, opened 2021-03-03
Unstructured scheme iris source

bjlittle:
This PR provides `iris` source GitHub artifact support, to allow testing against a specific `iris` GitHub repo branch name, tag name or commit sha.

It also contains condidtional task execution, better caching, and controls for conda environment verbosity.

The `flake8` and `black` linting tasks have been combined, and the long/variable running `osx` task as been retired.

PR #37, opened 2021-03-09
Unstructured scheme iris source (#33)

bjlittle:
This PR cherry-picks commit `be03823f7515f11543800dd39987f449a3c0aad3` from the `unstructured_scheme` feature branch and applies its goodness to `main`.

The benefit of this is that there is no need to wait for the feature branch to be merged back to `main` to use this functionality in `main`.

See #33

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -11,8 +11,13 @@
 #
 container:
   image: python:3.8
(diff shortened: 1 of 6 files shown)


stephenworsley (approved):
Looks good.

PR #39, opened 2021-03-12
Unstructured Scheme - Cube Creation 2D

stephenworsley:
Fleshes out the creation of cubes. Now adds grid coordinates from the target cube and copies over metadata and scalar coords from the source cube. Code is mostly copied from a similar iris functionality here https://github.com/SciTools/iris/blob/master/lib/iris/analysis/_regrid.py#L360-L363.

### DIFF
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -1,5 +1,7 @@
 """Provides an iris interface for unstructured regridding."""
 
+import copy
+
 import iris
 from iris.analysis._interpolation import get_xy_coords
 import numpy as np
@@ -86,7 +88,13 @@ def _create_cube(data, src_cube, mesh_dim, mesh, grid_x, grid_y):
 
     new_cube = iris.cube.Cube(data)
 
-    # TODO: add coords and metadata.
+    new_cube.add_dim_coord(grid_x, mesh_dim + 1)
+    new_cube.add_dim_coord(grid_y, mesh_dim)
+
+    new_cube.metadata = copy.deepcopy(src_cube.metadata)
+
+    for coord in src_cube.coords(dimensions=()):
+        new_cube.add_aux_coord(coord.copy())
 
     return new_cube
 
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -91,6 +91,11 @@ def _create_cube(data, src_cube, mesh_dim, grid_x, grid_y):
     # TODO: The following code assumes a 1D source cube and mesh_dim = 0.
     #  This is therefore simple code which should be updated when we start
     #  supporting the regridding of extra dimensions.
+
+    # TODO: The following code is rigid with respect to which dimensions
+    #  the x coord and y coord are assigned to. We should decide if it is
+    #  appropriate to copy the dimension ordering from the target cube
+    #  instead.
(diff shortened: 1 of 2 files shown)


esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -81,7 +83,13 @@ def _create_cube(data, src_cube, mesh_dim, mesh, grid_x, grid_y):
...
     new_cube = iris.cube.Cube(data)
 
-    # TODO: add coords and metadata.
+    new_cube.add_dim_coord(grid_x, mesh_dim + 1)
+    new_cube.add_dim_coord(grid_y, mesh_dim)
+
+    new_cube.metadata = copy.deepcopy(src_cube.metadata)
+
+    for coord in src_cube.coords(dimensions=()):
+        new_cube.add_aux_coord(coord.copy())
```
stephenworsley, line 92: Code is copied from https://github.com/SciTools/iris/blob/f76288cb82eefc5bc3c74461bfd8f397618eda14/lib/iris/analysis/_regrid.py#L360-L363 . This is from a regridder which is specifically for 2D cubes.

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -70,7 +72,7 @@ def _cube_to_GridInfo(cube):
 #     return result
 
 
-def _create_cube(data, src_cube, mesh_dim, mesh, grid_x, grid_y):
+def _create_cube(data, src_cube, mesh_dim, grid_x, grid_y):
```
stephenworsley, line 75: Originally this followed a pattern seen for example [here](https://github.com/SciTools/iris/blob/f76288cb82eefc5bc3c74461bfd8f397618eda14/lib/iris/analysis/_regrid.py#L819). In this pattern, (`x_dim`, `y_dim`) is equivalent to `mesh_dim` and (`src_x_coord`, `src_y_coord`) is equivalent to `mesh`. If we followed the same pattern then the purpose of `mesh` would be to determine which coordinates to copy over from `src_cube` to the new cube, this would be done by performing a comparison which used `mesh`. We anticipate that this would be done in multidimensional regridding. However, since a cube can only have one mesh, all that will need to be checked is whether the coord is a MeshCoord (or has a mesh attribute). It is therefore unlikely we will need `mesh` during cube creation and so I am removing it from the arguments.

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -81,7 +83,13 @@ def _create_cube(data, src_cube, mesh_dim, mesh, grid_x, grid_y):
 
     new_cube = iris.cube.Cube(data)
 
-    # TODO: add coords and metadata.
+    new_cube.add_dim_coord(grid_x, mesh_dim + 1)
+    new_cube.add_dim_coord(grid_y, mesh_dim)
```
stephenworsley, line 87: The dimension ordering is set so that `grid_y` is explicitly before `grid_x`. This will be the case even if the dimension ordering is the other way around in the target cube. The ability to be preserve the target cube dimension ordering may be added in future.

jamesp: @stephenworsley  thanks for making the changes we discussed. could you make this comment just a tad more descriptive of what would need to be done?  i.e. this is the specific case that will work for 2D, but needs to be reworked for 3D.  Thanks!

jamesp (approved):
Great work @stephenworsley , LGTM!

PR #46, opened 2021-03-26
Unstructured Scheme - Mesh Fetching

stephenworsley:
This PR adds support for fetching meshes and mesh dimensions from iris. Originally we had planned on using a helper function `_get_mesh_and_dim` to do this, however iris has updated to the point where this functionality is given in one line, so for clarities sake I have replaced the helper function. Instead of testing the helper function, I have tested `_regrid_unstructured_to_rectilinear__prepare`, where `_get_mesh_and_dim` would have been called. I have also added a test for `regrid_unstructured_to_rectilinear` to show that the whole regridder now functions as a result. Full test coverage can be added in a further PR.

### DIFF
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -176,7 +176,7 @@ class MeshToGridESMFRegridder:
         )
 
         # Store regrid info.
-        _, _, self.grid_x, self.grid_y, self.regridder = partial_regrid_info
+        _, self.grid_x, self.grid_y, self.regridder = partial_regrid_info
 
     def __call__(self, cube):
         """TODO: write docstring."""
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -110,6 +110,8 @@ def _regrid_unstructured_to_rectilinear__prepare(src_mesh_cube, target_grid_cube
 
     grid_x, grid_y = get_xy_dim_coords(target_grid_cube)
     mesh = src_mesh_cube.mesh
+    # TODO: Improve the checking of mesh validity. Check the mesh location and
+    #  raise appropriate error messages.
     assert mesh is not None
     # From src_mesh_cube, fetch the mesh, and the dimension on the cube which that
     # mesh belongs to.
@@ -180,9 +182,9 @@ class MeshToGridESMFRegridder:
 
     def __call__(self, cube):
         """TODO: write docstring."""
-        # TODO: Ensure cube has the same mesh as that of the recorded mesh.
-
         mesh = cube.mesh
+        # TODO: Ensure cube has the same mesh as that of the recorded mesh.
+        #  For the time being, we simply check that the mesh exists.
         assert mesh is not None
         mesh_dim = cube.mesh_dim()
 
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -3,7 +3,7 @@
 import copy
 
 import iris
-from iris.analysis._interpolation import get_xy_coords
+from iris.analysis._interpolation import get_xy_dim_coords
 import numpy as np
 
 # from numpy import ma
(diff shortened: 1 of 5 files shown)


esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -117,10 +108,12 @@ def _regrid_unstructured_to_rectilinear__prepare(src_mesh_cube, target_grid_cube
 
     # TODO: Record appropriate dimensions (i.e. which dimension the mesh belongs to)
 
-    grid_x, grid_y = get_xy_coords(target_grid_cube)
+    grid_x, grid_y = get_xy_dim_coords(target_grid_cube)
+    mesh = src_mesh_cube.mesh
+    assert mesh is not None
```
jamesp, line 113: Is this to be properly handled / exception raised in the future?  If so maybe a TODO added?

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -183,16 +176,17 @@ def __init__(self, src_mesh_cube, target_grid_cube, mdtol=1):
...
+        _, self.grid_x, self.grid_y, self.regridder = partial_regrid_info
 
     def __call__(self, cube):
         """TODO: write docstring."""
         # TODO: Ensure cube has the same mesh as that of the recorded mesh.
 
-        # mesh is probably an iris Mesh object, though it could also be a MeshCoord
-        mesh, mesh_dim = _get_mesh_and_dim(cube)
+        mesh = cube.mesh
+        assert mesh is not None
```
jamesp, line 186: TODO here if we plan on adding more error handling

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -117,10 +108,12 @@ def _regrid_unstructured_to_rectilinear__prepare(src_mesh_cube, target_grid_cube
...
 
-    grid_x, grid_y = get_xy_coords(target_grid_cube)
+    grid_x, grid_y = get_xy_dim_coords(target_grid_cube)
+    mesh = src_mesh_cube.mesh
+    assert mesh is not None
     # From src_mesh_cube, fetch the mesh, and the dimension on the cube which that
-    # mesh belongs to (mesh_dim).
-    mesh, mesh_dim = _get_mesh_and_dim(src_mesh_cube)
+    # mesh belongs to.
+    mesh_dim = src_mesh_cube.mesh_dim()
```
jamesp, line 116: Previous pseudocode version of this (L28) took `mesh` as a parameter. Is it not necessary to do this anymore?  Is there a possibility that a cube contains more than one mesh?

esmf_regrid/tests/unit/experimental/unstructured_scheme/test__regrid_unstructured_to_rectilinear__prepare.py
```diff
@@ -0,0 +1,65 @@
...
+    # These are not used in calculations and are here given a value of zero.
+    mesh_length = mesh.connectivity(contains_face=True).shape[0]
+    dummy_face_lon = AuxCoord(np.zeros(mesh_length), standard_name="longitude")
+    dummy_face_lat = AuxCoord(np.zeros(mesh_length), standard_name="latitude")
+    mesh.add_coords(face_x=dummy_face_lon, face_y=dummy_face_lat)
+    mesh.long_name = "example mesh"
+    return mesh
+
+
+def _flat_mesh_cube():
```
jamesp, line 33: Can you document what the object that this returns looks like? i.e. data is all ones?

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_regrid_unstructured_to_rectilinear.py
```diff
@@ -0,0 +1,54 @@
...
+        scalar_height = AuxCoord([5], units="m", standard_name="height")
+        scalar_time = DimCoord([10], units="s", standard_name="time")
+        result.add_aux_coord(scalar_height)
+        result.add_aux_coord(scalar_time)
+        return result
+
+    src = _add_metadata(src)
+    result = regrid_unstructured_to_rectilinear(src, tgt)
+
+    expected_data = np.ones([5, 6])
```
jamesp, line 46: It's not obvious to me why the expected data is [5,6] from the regrid operation, the test is relying on setup from another module in `_flat_mesh_cube`.  Could you add a comment explaining why this is the expected outcome?

jamesp (commented):
Looking good @stephenworsley 👍 .  A few comments inline

esmf_regrid/experimental/unstructured_scheme.py
stephenworsley: The previous pseudocode version was written before this functionality was implemented in iris. We've discussed the idea of a cube containing more than one mesh and we don't anticipate this possibility. In fact I don't believe it's possible to define such a thing under the current UGRID conventions.

PR #47, opened 2021-03-30
Unstructured Scheme - Cube Creation 3D

stephenworsley:
Expands on #39 by adding the ability to handle cubes with additional dimensions. Specifically, this copies from the source cube all DimCoords and AuxCoords which do not span the mesh dimension. Derived coordinates are a more complicated case which may be added in a future PR.

### DIFF
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -88,7 +88,10 @@ def _create_cube(data, src_cube, mesh_dim, grid_x, grid_y):
 
     new_cube.metadata = copy.deepcopy(src_cube.metadata)
 
-    coord_mapping = {}
+    # TODO: Handle derived coordinates. The following code is taken from
+    #  iris, the parts dealing with derived coordinates have been
+    #  commented out for the time being.
+    # coord_mapping = {}
 
     def copy_coords(src_coords, add_method):
         for coord in src_coords:
@@ -98,21 +101,21 @@ def _create_cube(data, src_cube, mesh_dim, grid_x, grid_y):
             dims = [dim if dim < mesh_dim else dim + 1 for dim in dims]
             result_coord = coord.copy()
             add_method(result_coord, dims)
-            coord_mapping[id(coord)] = result_coord
+            # coord_mapping[id(coord)] = result_coord
 
     copy_coords(src_cube.dim_coords, new_cube.add_dim_coord)
     copy_coords(src_cube.aux_coords, new_cube.add_aux_coord)
 
-    for factory in src_cube.aux_factories:
-        # TODO: Regrid dependant coordinates which span mesh_dim.
-        try:
-            result.add_aux_factory(factory.updated(coord_mapping))
-        except KeyError:
-            msg = (
-                "Cannot update aux_factory {!r} because of dropped"
-                " coordinates.".format(factory.name())
-            )
-            warnings.warn(msg)
+    # for factory in src_cube.aux_factories:
+    #     # TODO: Regrid dependant coordinates which span mesh_dim.
+    #     try:
+    #         result.add_aux_factory(factory.updated(coord_mapping))
+    #     except KeyError:
+    #         msg = (
+    #             "Cannot update aux_factory {!r} because of dropped"
+    #             " coordinates.".format(factory.name())
+    #         )
+    #         warnings.warn(msg)
 
     return new_cube
 
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -79,10 +79,6 @@ def _create_cube(data, src_cube, mesh_dim, grid_x, grid_y):
 
     new_cube = iris.cube.Cube(data)
 
-    # TODO: The following code assumes a 1D source cube and mesh_dim = 0.
-    #  This is therefore simple code which should be updated when we start
-    #  supporting the regridding of extra dimensions.
-
     # TODO: The following code is rigid with respect to which dimensions
     #  the x coord and y coord are assigned to. We should decide if it is
     #  appropriate to copy the dimension ordering from the target cube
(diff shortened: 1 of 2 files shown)


esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -101,8 +97,34 @@ def _create_cube(data, src_cube, mesh_dim, grid_x, grid_y):
...
+    #  iris, the parts dealing with derived coordinates have been
+    #  commented out for the time being.
+    # coord_mapping = {}
+
+    def copy_coords(src_coords, add_method):
+        for coord in src_coords:
+            dims = src_cube.coord_dims(coord)
+            if hasattr(coord, "mesh") or mesh_dim in dims:
+                continue
+            dims = [dim if dim < mesh_dim else dim + 1 for dim in dims]
```
jamesp, line 110: Can we have a comment to explain this line?  
I think from the `continue` statement above, `dim` will never equal `mesh_dim`, but explaining why it's going to be necessary to shift all later dimensions by 1 would make it clearer.

jamesp: If this code has been copied from iris then this comment probably belongs there too!

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -101,8 +97,34 @@ def _create_cube(data, src_cube, mesh_dim, grid_x, grid_y):
...
+            result_coord = coord.copy()
+            add_method(result_coord, dims)
```
jamesp, line 112: I *think* I get this, but slightly confusing as we're not returning anything from this function.  Is this because `add_method` is always going to be a method on a cube to which we are now appending these new coordinates, so the state change happens on the `add_method` owner?

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -88,10 +88,6 @@ def _create_cube(data, src_cube, mesh_dim, grid_x, grid_y):
```
jamesp, line 88: See Line 85: describes a `mesh` object, but this is no longer part of the `_create_cube` signature

esmf_regrid/tests/unit/experimental/unstructured_scheme/test__create_cube.py
```diff
@@ -11,17 +13,56 @@ def test_create_cube_2D():
...
     src_cube.add_aux_coord(scalar_height)
     src_cube.add_aux_coord(scalar_time)
 
     mesh_dim = 0
 
+    grid_x = DimCoord(np.arange(3), standard_name="longitude")
+    grid_y = DimCoord(np.arange(2), standard_name="latitude")
+
+    cube = _create_cube(data, src_cube, mesh_dim, grid_x, grid_y)
+    src_metadata = src_cube.metadata
```
jamesp, line 31: should `_create_cube` be handling metadata as well?

esmf_regrid/tests/unit/experimental/unstructured_scheme/test__create_cube.py
```diff
@@ -11,17 +13,56 @@ def test_create_cube_2D():
...
+    """Test creation of 2D output grid."""
+    data = np.ones([4, 2, 3, 5])
+
+    # Create a source cube with metadata and scalar coords
+    src_cube = Cube(np.zeros([4, 5, 5]))
+    src_cube.units = "K"
+    src_cube.attributes = {"a": 1}
+    src_cube.standard_name = "air_temperature"
+    scalar_height = AuxCoord([5], units="m", standard_name="height")
+    scalar_time = DimCoord([10], units="s", standard_name="time")
```
jamesp, line 52: Is this right? Adding a `DimCoord` as an `add_aux_coord`?

jamesp (commented):
Looking good @stephenworsley , a few comments inline

esmf_regrid/tests/unit/experimental/unstructured_scheme/test__create_cube.py
stephenworsley: This is a bit of Iris weirdness. While DimCoord and AuxCoord are different classes, a cube also has two "boxes" to put coords in: `_aux_coords_and_dims` and `_dim_coords_and_dims`. Somewhat counterintuitively, `_aux_coords_and_dims` can contain DimCoords. In fact, since this is a scalar coordinate, it has to be added to `_aux_coords_and_dims` since it does not describe a dimension (all coords in `_dim_coords_and_dims` must be associated with a dimension). Scalar DimCoords will be generated when a cube is sliced, so these are worth considering and testing with.

esmf_regrid/tests/unit/experimental/unstructured_scheme/test__create_cube.py
stephenworsley: `_create_cube` already handles metadata, this is also the case in iris.

jamesp: Ah yes, sorry I misread this section.  You're getting the source metadata to attach to the `expected_cube`, not to `cube`.

esmf_regrid/tests/unit/experimental/unstructured_scheme/test__create_cube.py
jamesp: Great, thanks for the explanation.  With that in mind, are we comfortable these tests are sufficiently complete in their coverage for now?

esmf_regrid/experimental/unstructured_scheme.py
stephenworsley: This line is new.

esmf_regrid/tests/unit/experimental/unstructured_scheme/test__create_cube.py
stephenworsley: I think so, there's still some stuff like derived coordinates which we might want to add in the future, but I think this covers the main stuff.

PR #55, opened 2021-04-13
Unstructured Scheme - Extra Dims

stephenworsley:
Provide support for cubes with extra dimensions. This change allows the regridding of cubes whose `mesh_dim` is any dimension. This is done by transposing the data before and after the regridding opperation. In future, this may be handled by the regridder itself (as opposed to the regridding scheme), however, keeping this functionality in a helper function allows the feature branch to stay more closely aligned with main.

### DIFF
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -5,7 +5,6 @@ import copy
 import iris
 from iris.analysis._interpolation import get_xy_dim_coords
 import numpy as np
-from numpy import ma
 
 from esmf_regrid.esmf_regridder import GridInfo, Regridder
 from esmf_regrid.experimental.unstructured_regrid import MeshInfo
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -63,8 +63,7 @@ def _regrid_along_dims(regridder, data, src_dim, mdtol):
     result = regridder.regrid(data, mdtol=mdtol)
 
     # Move grid axes back into the original position of the mesh.
-    result = np.moveaxis(result, -1, src_dim)
-    result = np.moveaxis(result, -1, src_dim)
+    result = np.moveaxis(result, [-2, -1], [src_dim, src_dim + 1])
 
     return result
 
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -57,17 +57,15 @@ def _cube_to_GridInfo(cube):
 def _regrid_along_dims(regridder, data, src_dim, mdtol):
     # Before regridding, data is transposed to a standard form.
     # In the future, this may be done within the regridder by specifying args.
-    data_dims = len(data.shape)
-    new_axes_in = list(range(data_dims))
-    new_axes_in.pop(src_dim)
-    new_axes_in.append(src_dim)
-    data = ma.transpose(data, axes=new_axes_in)
+
+    # Move the mesh axis to be the last dimension.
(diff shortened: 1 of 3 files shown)


esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -55,17 +53,19 @@ def _cube_to_GridInfo(cube):
...
-#     # This will be done either with something like the following code
-#     # or else done within the regridder by specifying args.
-#     # new_axes = list(range(len(data.shape)))
-#     # new_axes.pop(src_dim)
-#     # new_axes.append(src_dim)
-#     # data = ma.transpose(data, axes=new_axes)
-#
-#     result = regridder.regrid(data, mdtol=mdtol)
-#     return result
+def _regrid_along_dims(regridder, data, src_dim, mdtol):
```
jamesp, line 56: Is the name of this function as clear as it could be?  what do we mean by "along dims"?  is `src_dim` the same as `mesh_dim` name used elsewhere?

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_MeshToGridESMFRegridder.py
```diff
@@ -59,6 +61,54 @@ def _add_metadata(cube):
...
+    tgt = _grid_cube(n_lons, n_lats, lon_bounds, lat_bounds, circular=True)
+
+    src_cube = mesh_cube.copy()
+    src_cube.transpose([1, 0, 2])
+    regridder = MeshToGridESMFRegridder(src_cube, tgt)
+    result = regridder(mesh_cube)
+
+    # Lenient check for data.
+    expected_data = np.zeros([t, n_lats, n_lons, h])
+    assert np.allclose(expected_data, result.data)
```
jamesp, line 99: Would it be better to do these tests with ones rather than zeros?  Would we catch more errors?

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_regrid_unstructured_to_rectilinear.py
```diff
@@ -55,3 +57,51 @@ def _add_metadata(cube):
...
+    n_lats = 5
+    lon_bounds = (-180, 180)
+    lat_bounds = (-90, 90)
+    tgt = _grid_cube(n_lons, n_lats, lon_bounds, lat_bounds, circular=True)
+
+    result = regrid_unstructured_to_rectilinear(cube, tgt)
+
+    # Lenient check for data.
+    expected_data = np.empty([t, n_lats, n_lons, h])
+    expected_data[:] = np.arange(t * h).reshape(t, h)[:, np.newaxis, np.newaxis, :]
```
jamesp, line 96: What is the purpose of the empty -> fill with arange construction?

This test and the `MeshToGridESMFRegridder` test are very similar, is there a reason we're using different test data for each?  Should we have a test that compares both the object and the function return the same thing?

jamesp (commented):
Looking good 👍 , a few comments inline

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_regrid_unstructured_to_rectilinear.py
stephenworsley: The idea here is that each horizontal level is a unique constant. I do this by broadcasting an array filled with different numbers from `np.arange`.
The purpose here is to ensure that each horizontal level is mapped to the appropriate place in the final array. Having each level being constant makes this straightforward to check that this calculation is correct. I've changed the other test in `MeshToGridESMFRegridder` to be equivalent.

jamesp (approved):

PR #68, opened 2021-05-18
Conda lock

trexfeathers:
Mirrors https://github.com/SciTools/iris/pull/4108, for iris-esmf-regrid. This may seem like overkill, but this repo is optionally dependent on Iris and that brings a larger set of dependencies and makes resolution issues more likely.

If this is merged I will immediately create another PR targeted at the [`unstructured_scheme` feature branch](https://github.com/SciTools-incubator/iris-esmf-regrid/tree/unstructured_scheme), with appropriate lock files for that branch.

### Benefits

* Faster environment creation time.
  Especially valuable for CI.
* Known issues with `conda env update` no longer threaten CI runs.
* Keeps dependency changes isolated - "atomic changes" - no longer have to bundle fixes into other PR's unexpectedly.
* Traceable dependency changes at specific, dedicated commits.
  Particularly useful for future performance testing plans.

### Maintenance

* Every week a PR against `main` will be automatically generated, proposing the dependency updates.
  * If there are no changes, no PR is generated.
  * This can be merged, rejected, or work can be done on the PR to resolve test failures.
  * Here's an example from an earlier test: trexfeathers#2
* Feature branches will need manual updating.
  * Safest, since we can't predict the purpose of a feature branch.
  * Done by running: `nox --session update_lockfiles` and pushing the changed files.

@jamesp: please could you give this a once-over? I deviated slightly in my implementation, particularly with making a Nox session specifically for updating the lockfiles, so I won't to make sure I haven't made any big mistakes. It seems to test OK. Thanks!


trexfeathers:
Also of note: I've made use of [Nox's `Session.run_always()`](https://nox.thea.codes/en/stable/config.html#nox.sessions.Session.run_always) in several places, so that we can  make effective use of [Nox's `--install-only` argument](https://nox.thea.codes/en/stable/usage.html#skipping-everything-but-install-commands) in future - part of my plans for performance testing with ASV.

noxfile.py
```diff
@@ -30,68 +33,76 @@
...
+def _venv_populated(session: nox.sessions.Session) -> bool:
+    """Return True if the Conda venv has been created and the list of packages in the lockfile installed."""
+    return _session_cachefile(session).is_file()
+
+
+def _venv_changed(session: nox.sessions.Session) -> bool:
+    """Return True if the installed session is different to that specified in the lockfile."""
+    result = False
+    if _venv_populated(session):
+        with _session_lockfile(session).open("rb") as lockfile:
```
jamesp, line 78: I suppose a simpler alternative here might just be to cache the actual lockfile, rather than a hash of the lockfile?  It would also allow us to do introspection on the differences if we ever wanted to

jamesp: (I know this is the same as I did for iris and was based on the previous caching technique, it's now struck me that this might not be the optimal solution!)

noxfile.py
```diff
@@ -130,34 +149,129 @@ def _get_iris_github_artifact(session):
...
+        session.run_always(
+            "conda",
+            "list",
+            f"--prefix={venv_dir}",
+            "--explicit",
+        )
+
+
+@nox.session
+def update_lockfiles(session: nox.sessions.Session):
```
jamesp, line 207: Nice addition

noxfile.py
trexfeathers: Nice, I'll get on that

trexfeathers: e800e5b7281e00984910268cebecb68f2f735f41

.github/workflows/refresh-lockfiles.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: install nox
+        run: |
+          source $CONDA/bin/activate base
+          conda install -y -c conda-forge nox pyyaml
+      - name: generate lockfiles
+        run: $CONDA/bin/nox --session update_lockfiles
+      - name: create pull request
+        uses: peter-evans/create-pull-request@052fc72b4198ba9fbc81b818c6e1859f747d49a8
```
stephenworsley, line 35: I was wondering if there are any security implications to pointing to someone elses GitHub repo. I'm guessing that this is more safe because its pointing to a specific commit, but I was wondering how far you had considered this.

trexfeathers: You're more on it than I was when I first used GHA's - I had it pointing to a tag rather than a commit! Pointing to a specific commit is a recognised way of alleviating this security risk, [see here](https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions).

zklaus:
@valeriupredoi, this might be of interest for ESMValTool as well. As I understand it, you solve the conda environment once per week and then lock every version, completely avoiding the solving step in the regular CI.

valeriupredoi:
cheers @zklaus - apologies @trexfeathers for crashing your PR :grin: This is a nice feature that I didn't know about but I am wondering how much different is it from creating an environment from a saved explicit environment export (seeing that the lockfile is actually a list of sources from a conda env export). On the other hand, this could be very useful as a lifeboat for CI testing and user source installs if the instantaneous environment fails to solve. Klaus, I would argue we should have this in place but only if the env solving fails, otherwise not relying on a weekly lock on a regular base :+1:

trexfeathers:
>I am wondering how much different is it from creating an environment from a saved explicit environment export

@valeriupredoi it's exactly the same, except that it skips the environment creation altogether and _just_ creates the export.

jamesp:
Under the covers conda-lock uses `conda create --dry-run` to generate the explicit list of packages so in theory(!) you should get exactly the same outcome regardless of method.

valeriupredoi:
good stuff, gents! I have already opened an issue in our repo get the discussion and eventual implementation going :beer:

stephenworsley (approved):
Looks good to me.

PR #70, opened 2021-05-19
Abstract base class

### LINKED ISSUE #45: Introduce common (abstract?) base class for MeshInfo and GridInfo
This will allow us to share some common code, namely
```
@property
def dims(self):
    return len(self.shape)

@property
def size(self):
    return prod(self.shape)
```
as well as `_flatten_array` and `_unflatten_array`.

zklaus:
This introduces a common base class for GridInfo and MeshInfo. Additionally, it brings minor refactorings.

Closes #45


esmf_regrid/_esmf_sdo.py
```diff
@@ -0,0 +1,225 @@
...
+        self._shape = shape
+        self._index_offset = index_offset
+        self._field_kwargs = field_kwargs
+
+    @abstractmethod
+    def _make_esmf_sdo(self):
+        pass
+
+    def make_esmf_field(self):
+        """Return an ESMF field representing the grid."""
```
stephenworsley, line 29: References to "grid" ought to be replaced, perhaps with either "object" or "spatial discretisation object".

esmf_regrid/_esmf_sdo.py
```diff
@@ -0,0 +1,225 @@
+# -*- coding: utf-8 -*-
```
stephenworsley, line 1: Is there a specific reason for this line? Is this worth adding to other files?

stephenworsley (changes requested):
Looks good to me, just a few very minor comments. Otherwise this looks good to go.

esmf_regrid/_esmf_sdo.py
zklaus: More of a habit. It did have its use in Python 2 and is still useful for my beloved emacs, but I just read up on it and it seems in Python 3 it is no longer expected because utf-8 already is the default. So I'll remove it for consistency.

stephenworsley (approved):
Looks good to me

Issue #64, opened 2021-05-07
Add grid-to-grid regridding scheme


stephenworsley:
## ✨ Feature Request

Add a regridding scheme which maps a rectilinear cube onto a rectilinear cube. This can initially follow the pattern established here :https://github.com/SciTools-incubator/iris-esmf-regrid/blob/unstructured_scheme/esmf_regrid/experimental/unstructured_scheme.py (with a view to incorporating ideas from #26). Coppying this pattern is probably the quickest way to achieve a functional regridding scheme which can then be expanded to flesh out and restructure the scheme, incorporating ideas from #26 for example.

jamesp:
@stephenworsley I think we can close this issue as done, right?

stephenworsley:
@jamesp yeah, I guess this is done now. Closed by #75.

PR #75, opened 2021-05-24
Rectilinear scheme

stephenworsley:
Addresses #64.


esmf_regrid/schemes.py
```diff
@@ -1,19 +1,315 @@
...
+        src_grid : cube
+            The cube defining the source grid.
+        tgt_grid : cube
+            The cube defining the target grid.
+
+        Returns
+        -------
+        ESMFAreaWeightedRegridder
+            A callable with the interface:
+                `callable(cube)`
```
jamesp, line 241: `callable` is a built-in function in python, can we write this to be more explicit?

I assume you mean something like 
> Returns a ESMFAreaWeightedRegridder regridder object, that can be called `regridder(cube)` where `cube` is a ...

esmf_regrid/schemes.py
```diff
@@ -1,19 +1,315 @@
...
+        _bounds_cf_to_simple_1d(lon.bounds),
+        _bounds_cf_to_simple_1d(lat.bounds),
+        crs=crs,
+        circular=lon.circular,
+    )
+
+
+def _regrid_along_grid_dims(regridder, data, grid_x_dim, grid_y_dim, mdtol):
+    data = np.moveaxis(data, [grid_x_dim, grid_y_dim], [-1, -2])
+    result = regridder.regrid(data, mdtol=mdtol)
```
jamesp, line 56: This seems to be a different interface to the `regridder(data)` below.  Is the `regridder` argument a `Regridder` or an `ESMFWeightedAreaRegridder`? This might be the perfect opportunity to add some type hints!

esmf_regrid/schemes.py
```diff
@@ -1,19 +1,315 @@
...
+
+    grid_x_dim = src_grid_cube.coord_dims(src_x)[0]
+    grid_y_dim = src_grid_cube.coord_dims(src_y)[0]
+
+    srcinfo = _cube_to_GridInfo(src_grid_cube)
+    tgtinfo = _cube_to_GridInfo(tgt_grid_cube)
+
+    regridder = Regridder(srcinfo, tgtinfo)
+
+    regrid_info = (grid_x_dim, grid_y_dim, tgt_x, tgt_y, regridder)
```
jamesp, line 131: probably worth considering making `regrid_info` a named tuple if it is going to be passed around between functions.  It will make it much easier to debug if we want to add more values to it in the future

esmf_regrid/schemes.py
jamesp: Incidentally, why do these two objects have different interfaces? From the names you'd intuitively expect one to be a subclass of the other, but that is not the case.

esmf_regrid/schemes.py
```diff
@@ -1,19 +1,315 @@
...
+import iris
+from iris.analysis._interpolation import get_xy_dim_coords
+import numpy as np
+
+from esmf_regrid.esmf_regridder import GridInfo, Regridder
 
 __all__ = [
     "ESMFAreaWeighted",
+    "ESMFAreaWeightedRegridder",
+    "regrid_rectilinear_to_rectilinear",
```
jamesp, line 14: Is this intended to be the final name / export of this library, or will the final regridder class abstract away the differences between passing in rectilinear or unstructured data?

esmf_regrid/tests/unit/schemes/test_ESMFAreaWeightedRegridder.py
```diff
@@ -0,0 +1,81 @@
...
+
+    n_lons_dif = 7
+    src_dif_coord = _grid_cube(
+        n_lons_dif, n_lats, lon_bounds, lat_bounds, circular=True
+    )
+    src_dif_circ = _grid_cube(n_lons, n_lats, lon_bounds, lat_bounds, circular=False)
+
+    regridder = ESMFAreaWeightedRegridder(src_init, tgt)
+
+    with pytest.raises(AssertionError):
```
jamesp, line 59: Document here the expected assertion error, or include a further test for asserting it's hitting the error we're looking for https://docs.pytest.org/en/6.2.x/assert.html#assertions-about-expected-exceptions

esmf_regrid/tests/unit/schemes/test_ESMFAreaWeightedRegridder.py
```diff
@@ -0,0 +1,81 @@
...
+    Checks that an error is raised when mdtol is out of range.
+    """
+    n_lons = 6
+    n_lats = 5
+    lon_bounds = (-180, 180)
+    lat_bounds = (-90, 90)
+    src = _grid_cube(n_lons, n_lats, lon_bounds, lat_bounds, circular=True)
+    tgt = _grid_cube(n_lons, n_lats, lon_bounds, lat_bounds, circular=True)
+
+    with pytest.raises(ValueError):
```
jamesp, line 78: Same again, document (with comments) or include within the test the specific value error we're testing for

esmf_regrid/tests/unit/schemes/test_regrid_rectilinear_to_rectilinear.py
```diff
@@ -0,0 +1,150 @@
...
+    # Lenient check for data.
+    assert np.allclose(no_mdtol_expected_data, no_mdtol_result.data)
+    assert np.allclose(expected_data, full_mdtol_result.data)
+
+
+def test_extra_dims():
+    """
+    Test for :func:`esmf_regrid.schemes.regrid_rectilinear_to_rectilinear`.
+
+    Tests the handling of extra dimensions and metadata.
```
jamesp, line 71: Maybe expand on this a little bit more.  i.e. what are we expecting the handling to do?  Pass through all metadata and extra dims?  Expect the ordering of dims to be the same or different?

jamesp (commented):
Looks good @stephenworsley, some comments inline.

It's probably worth us at our next group meeting having a discussion about API - how do we want users to interact with the code, and is it consistent with iris and ESMF?  It looks like it is is not going to be consistent with everything so we'll have to make some choices.

requirements/py38.yml
```diff
@@ -14,6 +14,7 @@ dependencies:
   - numpy>=1.14
   - scipy
   - esmpy>=7.0
+  - iris
```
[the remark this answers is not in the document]
stephenworsley: The alternate version of iris is just for the feature branch. The reason the feature branch is still a feature branch is because it depends on an unreleased  version of iris. This scheme only works with non-mesh cubes so it can go on main and use the normal versio of iris.

esmf_regrid/schemes.py
stephenworsley: So the `ESMFWeightedAreaRegridder` deals with iris objects and `Regridder` deals with `GridInfo`/`MeshInfo` objects. For this reason, I don't think it's appropriate to have one be a subclass of the other. `Regridder` is more generic while `ESMFWeightedAreaRegridder` is dealing with more specific cases. There is probably a case for the names to be more thought out. I believe we came up with these names a while ago and at the time the name `ESMFWeightedAreaRegridder` suggested that what would be specific about this regridding scheme is that it would use ESMF and be an instance of area weighted regridding. I suppose the idea was that eventually our back end would be able to handle more kinds of regridding and potentially use different packages to do regridding with. At the present, we don't have any other kinds of regridding so this isn't really what sets this appart from `Regridder`, though this could still change in the future.

esmf_regrid/schemes.py
stephenworsley: That's fair, though it's probably worth noting that we're likely to end up with different regridders which will store different pieces of information. I think we would have to have a named tuple for each scheme.

esmf_regrid/schemes.py
stephenworsley: I think it's possible that ESMFAreaWeighted might be able to act in a way that is similar to the other two objects here. It is in a way the top level object which is why it was initially the only one included in `__all__`. With that said, I'm kind of going off my instructions for the unstructured case which involved presenting objects with interfaces like the two i have introduced here. I've included everything for the sake of completeness, though there is probably some redundancy here which I could imagine us tidying up in future.

jamesp (approved):

PR #79, opened 2021-06-02
CI Benchmarking

trexfeathers:
* Enable benchmark integration into CI.
* Post-merge full run including publishing to a `gh-pages` branch.
  If this ends up running too long we can review. Perhaps a smaller range, skipping commits, or something completely different.
* A fix for `pyproject.toml`, without which Black will run on all ASV environments as well as the source code.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -17,10 +17,8 @@ container:
 env:
   # Skip specific tasks by name. Set to a non-empty string to skip.
   SKIP_LINT_TASK: ""
-  SKIP_TEST_TASK: ""
-  # Currently disabled:
-  #  https://github.com/SciTools-incubator/iris-esmf-regrid/pull/76#issuecomment-850378861
-  SKIP_BENCHMARK_TASK: "."
+  SKIP_TEST_TASK: "."
+  SKIP_BENCHMARK_TASK: ""
   # Maximum cache period (in weeks) before forcing a new cache upload.
   CACHE_PERIOD: "2"
   # Increment the build number to force new conda cache upload.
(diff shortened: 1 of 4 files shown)


.cirrus.yml
```diff
@@ -118,6 +116,17 @@ benchmark_task:
...
   name: "${CIRRUS_OS}: performance benchmarking"
+  # Custom clone behaviour to enable ASV to access the PR base branch (if on a
+  #  PR).
+  clone_script: |
+    if [ -z "$CIRRUS_PR" ]; then
+      git clone --branch=$CIRRUS_BRANCH https://x-access-token:${CIRRUS_REPO_CLONE_TOKEN}@github.com/${CIRRUS_REPO_FULL_NAME}.git $CIRRUS_WORKING_DIR
+      git reset --hard $CIRRUS_CHANGE_IN_REPO
+    else
+      git clone --recursive https://x-access-token:${CIRRUS_REPO_CLONE_TOKEN}@github.com/${CIRRUS_REPO_FULL_NAME}.git $CIRRUS_WORKING_DIR
+      git fetch origin pull/$CIRRUS_PR/head:pull/$CIRRUS_PR
```
stephenworsley, line 127: Should `CIRRUS_PR` not be in curly brackets here?

stephenworsley (changes requested):
Looks good, just a small style question.

.cirrus.yml
trexfeathers: Since our use case allowed me to borrow wholesale from [Cirrus' own example](https://cirrus-ci.org/guide/tips-and-tricks/#custom-clone-command), I never thought about that. Certainly looks like it should be as you say. We can't really test whether it works until it's merged!

ff01df5b6da60850d8dcc8ee5f5813c8e314fa0b

stephenworsley (approved):

Issue #61, opened 2021-04-21
Add real integration and gradient field testing


abooton:
## ✨ Feature Request
Add integration testing so that KGO is routinely tested

Integration tests are needed for the following:
* Real data:  From the [iris-test-data repository](https://github.com/SciTools/iris-test-data/tree/master/test_data/NetCDF):
   Testing the regridding of: Tripolar (ORCA) grid, Lambert-Azimutal-Equal-Area, Rotated pole, and ugrid data.  (Tripolar is the most important).

* Gradient field data:  Should enable regidding of different resolutions to be easily assessed

KGO:
* The known-good-output should be checking the exact data values (rather than visualisation of the regridding).
- This could be implemented in line with the methods currently used in Iris.

jamesp:
not stale

PR #80, opened 2021-06-02
Support lazy regridding

stephenworsley:
Provides lazy regridding through essentially the same means as in iris (introduced in this PR https://github.com/SciTools/iris/pull/3701).

### DIFF
--- a/benchmarks/benchmarks/ci/esmf_regridder.py
+++ b/benchmarks/benchmarks/ci/esmf_regridder.py
@@ -150,4 +150,4 @@ class TimeLazyRegridding:
 
     def time_regridding_realisation(self, cache):
         assert self.result.has_lazy_data()
-        _ = result.data
+        _ = self.result.data
--- a/benchmarks/benchmarks/ci/esmf_regridder.py
+++ b/benchmarks/benchmarks/ci/esmf_regridder.py
@@ -112,7 +112,7 @@ class TimeLazyRegridding:
         n_lats_src = 200
         n_lons_tgt = 20
         n_lats_tgt = 40
-        h = 100
+        h = 2000
         # Rotated coord systems prevent pickling of the regridder so are
         # removed for the time being.
         # coord_system_src = RotatedGeogCS(0, 90, 90)
--- a/benchmarks/benchmarks/ci/esmf_regridder.py
+++ b/benchmarks/benchmarks/ci/esmf_regridder.py
@@ -138,10 +138,16 @@ class TimeLazyRegridding:
         return regridder, file
 
     def setup(self, cache):
-        _, file = cache
+        regridder, file = cache
         self.src = iris.load_cube(file)
+        cube = iris.load_cube(file)
+        self.result = regridder(cube)
 
     def time_lazy_regridding(self, cache):
         assert self.src.has_lazy_data()
         regridder, _ = cache
         _ = regridder(self.src)
+
+    def time_regridding_realisation(self, cache):
+        assert self.result.has_lazy_data()
+        _ = result.data
--- a/esmf_regrid/_esmf_sdo.py
+++ b/esmf_regrid/_esmf_sdo.py
@@ -61,7 +61,7 @@ class SDO(ABC):
         We then take the transpose so that matrix multiplication happens over
         the appropriate axes.
         """
-        return array.reshape(-1, self.size, order="F").T
+        return array.T.reshape(self.size, -1)
 
     def _matrix_to_array(self, array, extra_dims):
         """
@@ -69,7 +69,7 @@ class SDO(ABC):
 
         This is the inverse operation of `_array_to_matrix`.
         """
-        return array.T.reshape(extra_dims + self._shape, order="F")
+        return array.reshape((extra_dims + self._shape)[::-1]).T
 
 
 class GridInfo(SDO):
--- a/esmf_regrid/_esmf_sdo.py
+++ b/esmf_regrid/_esmf_sdo.py
@@ -61,7 +61,7 @@ class SDO(ABC):
         We then take the transpose so that matrix multiplication happens over
         the appropriate axes.
         """
-        return array.T.reshape(self.size, -1)
+        return array.T.reshape((self.size, -1))
 
     def _matrix_to_array(self, array, extra_dims):
         """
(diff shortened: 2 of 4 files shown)


esmf_regrid/_esmf_sdo.py
```diff
@@ -61,15 +61,15 @@ def _array_to_matrix(self, array):
         We then take the transpose so that matrix multiplication happens over
         the appropriate axes.
         """
-        return array.reshape(-1, self.size, order="F").T
+        return array.T.reshape(self.size, -1)
```
jamesp, line 64: This is functionally equivalent to the call on L72, but would be nice to be consistent in how we call the numpy api. I personally prefer the explicit tuple of shape integers on L72, as it is compatible with `np.reshape`.

See notes on https://numpy.org/doc/stable/reference/generated/numpy.ndarray.reshape.html#numpy.ndarray.reshape

esmf_regrid/schemes.py
```diff
@@ -2,8 +2,10 @@
 
 from collections import namedtuple
 import copy
+import functools
 
 import iris
+from iris._lazy_data import map_complete_blocks
```
jamesp, line 8: Is there any alternative to importing from a _ module (i.e. private) in iris? Not a deal breaker, these two libraries are very tightly coupled already.

esmf_regrid/tests/unit/schemes/test_regrid_rectilinear_to_rectilinear.py
```diff
@@ -149,3 +150,19 @@ def _add_metadata(cube):
...
+    lon_bounds = (-180, 180)
+    lat_bounds = (-90, 90)
+    src = _grid_cube(n_lons, n_lats, lon_bounds, lat_bounds, circular=True)
+    tgt = _grid_cube(n_lons, n_lats, lon_bounds, lat_bounds, circular=True)
+
+    src_data = da.ones([n_lats, n_lons])
+    src.data = src_data
+    assert src.has_lazy_data()
+    result = regrid_rectilinear_to_rectilinear(src, tgt)
+    assert result.has_lazy_data()
```
jamesp, line 168: Do we have a test that ensures that lazy regridding returns expected results after computing?

jamesp: Could be a parameterised test, e.g. pass in the same data, lazy and non lazy, and ensure they both give the same reference result

esmf_regrid/schemes.py
```diff
@@ -170,10 +172,19 @@ def _regrid_rectilinear_to_rectilinear__prepare(src_grid_cube, tgt_grid_cube):
...
+        regridder,
+        grid_x_dim=grid_x_dim,
+        grid_y_dim=grid_y_dim,
+        mdtol=mdtol,
+    )
+
+    new_data = map_complete_blocks(
+        src_cube,
+        regrid,
+        (grid_x_dim, grid_y_dim),
```
jamesp, line 186: A bit of documentation would be useful here - a comment to explain why you are passing the horizontal dims as non-chunkable to the map function.

jamesp (changes requested):
Looks a good, a couple of minor points inline

stephenworsley:
When this is merged, the performance test in #100 will be measuring the set up for lazy regridding but not the final realisation of the data. The realisation will also be worth measuring after this is merged in (I don't think it will make sense to add such a test before this is merged in since then it would be measuring the realisation of an already realised array).

esmf_regrid/schemes.py
stephenworsley: As far as I can tell, the alternative would be to effectively rewrite the function. I don't believe it's a very large or complicated function so it could be doable, but since this work is more or less inspired by the work on iris I've been keeping the links obvious for now. I imagine this may end up being replaced when optimising for performance eventually.

jamesp: No need to rewrite, this library is tightly coupled to iris so it makes sense to reuse the code over there.  We may want to move `map_complete_blocks` to the public api in iris if this becomes a regular thing we do.

stephenworsley:
The following excerpt from the benchmarking CI illustrates the effect these changes have on performance:
image
The actual regridding step now takes up a negligible amount of time. The regridding calculations are now done during realisation of the data (previously the data was already realised). This calculation step now takes almost exactly half the time it did before (probably due to dask parallelisation).

benchmarks/benchmarks/ci/esmf_regridder.py
```diff
@@ -112,7 +112,7 @@ def setup_cache(self):
         n_lats_src = 200
         n_lons_tgt = 20
         n_lats_tgt = 40
-        h = 100
+        h = 2000
```
stephenworsley, line 115: Size of the source cube is increased since iris has a minimum size of chunk when loading a cube. If the source cube isn't large enough, there will only be one chunk and parallelisation cannot be demonstrated.

esmf_regrid/schemes.py
stephenworsley: I've added a comment explaining.

esmf_regrid/tests/unit/schemes/test_regrid_rectilinear_to_rectilinear.py
stephenworsley: I've added a bit to flesh out the test.

jamesp: So what are you doing here? regridding to itself and ensuring no data has changed?

stephenworsley: Effectively, yes. With chunking added in all directions. This effectively makes sure that the dimensions aren't mixed up and everything stays in the correct order. I feel like this should cover most of the obvious bugs you might expect from dealing with chunked data.

stephenworsley: Though with that said, there are less obvious bugs such as this one https://github.com/dask/dask/issues/6907 which I imagine would not be picked up. This particular bug seems to be obscure enough that we probably wouldn't pick it up even with a more sophisticated test.

jamesp: Agree, seems reasonable.  We should keep track of #61 on the backlog though and make sure we get to it at some point as we can use a slightly more complex case there to ensure things are working as expected.

jamesp (approved):

esmf_regrid/tests/unit/schemes/test_regrid_rectilinear_to_rectilinear.py
stephenworsley: I suppose it could be an idea to have a regridding between different grids, but to compare the result from realised regridding to ensure it matches the result from lazy regridding. That way the test is closer to actual regridding use cases.

jamesp:
OK I'm happy with this. Any last changes @stephenworsley ?

PR #83, opened 2021-06-04
Add docstrings

stephenworsley:
Add docstrings which were still missing.


jamesp (changes requested):
See if you can fix the linting on this and we're good to go :+1:

jamesp (approved):

PR #92, opened 2021-06-28
Rename lockfile_dir to avoid builtin name

lbdreyer:
Noticed this small issue: in one of the noxfile functions, it names a variable `dir`. [dir is a built in function](https://docs.python.org/3/library/functions.html#dir) so it shouldn't be used to as a variable name, as it overrides the builtin function.


stephenworsley (approved):
Looks good.

PR #98, opened 2021-07-28
Add regridding benchmark

stephenworsley:
Adds benchmarks for the performing of regridding (after initialisation of a regridder). This is with a view to measuring the performance of different regridding solutions (i.e. sparse vs scipy.sparse). The benchmark is written to include cases involving higher to lower resolution, lower to higher, similar resolutions and a case where parts of the grid are high to low res and other parts are low to high res due to a rotated pole.

### DIFF
--- a/benchmarks/benchmarks/ci/esmf_regridder.py
+++ b/benchmarks/benchmarks/ci/esmf_regridder.py
@@ -72,7 +72,9 @@ class TimeRegridding:
             coord_system=coord_system_src,
         )
         tgt = _grid_cube(n_lons_tgt, n_lats_tgt, lon_bounds, lat_bounds)
-        src_data = np.arange(n_lats_src * n_lons_src * h).reshape([n_lats_src, n_lons_src, h])
+        src_data = np.arange(n_lats_src * n_lons_src * h).reshape(
+            [n_lats_src, n_lons_src, h]
+        )
         src = Cube(src_data)
         src.add_dim_coord(grid.coord("latitude"), 0)
         src.add_dim_coord(grid.coord("longitude"), 1)
--- a/benchmarks/benchmarks/ci/esmf_regridder.py
+++ b/benchmarks/benchmarks/ci/esmf_regridder.py
@@ -2,6 +2,7 @@
 
 import numpy as np
 from iris.coord_systems import RotatedGeogCS
+from iris.cube import Cube
 
 from esmf_regrid.esmf_regridder import GridInfo
 from esmf_regrid.schemes import ESMFAreaWeightedRegridder
@@ -52,6 +53,7 @@ class TimeRegridding:
         n_lats_src = 40
         n_lons_tgt = 20
         n_lats_tgt = 40
+        h = 100
         if type == "large source":
             n_lons_src = 100
             n_lats_src = 200
@@ -62,7 +64,7 @@ class TimeRegridding:
             coord_system_src = RotatedGeogCS(0, 90, 90)
         else:
             coord_system_src = None
-        src = _grid_cube(
+        grid = _grid_cube(
             n_lons_src,
             n_lats_src,
             lon_bounds,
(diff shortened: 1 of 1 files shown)


benchmarks/benchmarks/ci/esmf_regridder.py
```diff
@@ -37,3 +41,45 @@ def time_make_grid(self):
...
+        grid = _grid_cube(
+            n_lons_src,
+            n_lats_src,
+            lon_bounds,
+            lat_bounds,
+            coord_system=coord_system_src,
+        )
+        tgt = _grid_cube(n_lons_tgt, n_lats_tgt, lon_bounds, lat_bounds)
+        src_data = np.arange(n_lats_src * n_lons_src * h).reshape(
+            [n_lats_src, n_lons_src, h]
```
trexfeathers, line 76: I assume the switch in ordering from lon-lat to lat-lon is deliberate to make sure that some regridding work is required? Otherwise it's unnecessarily confusing.

benchmarks/benchmarks/ci/esmf_regridder.py
```diff
@@ -37,3 +41,45 @@ def time_make_grid(self):
         esmf_grid.data[:] = 0
 
     time_make_grid.version = 1
+
+
+class TimeRegridding:
+    params = ["similar", "large source", "large target", "mixed"]
```
trexfeathers, line 47: I'm not sure what name you might use, but are you aware that you could name this `param` by using the `param_names` variable? This makes things more readable when viewing published results.

trexfeathers (commented):
I think this looks good to me, although I'm inexperienced in the use of the package so I might have missed something.

benchmarks/benchmarks/ci/esmf_regridder.py
stephenworsley: I think this might be a hang over from somewhere in CF refering to dimensions in the reverse order to how they are described in numpy. I can't remember why it is this way, but i believe this is inherited from how it is done in other tests. I might have another look into if this still makes sense.

benchmarks/benchmarks/ci/esmf_regridder.py
```diff
@@ -37,3 +41,46 @@ def time_make_grid(self):
...
+        )
+        tgt = _grid_cube(n_lons_tgt, n_lats_tgt, lon_bounds, lat_bounds)
+        src_data = np.arange(n_lats_src * n_lons_src * h).reshape(
+            [n_lats_src, n_lons_src, h]
+        )
+        src = Cube(src_data)
+        src.add_dim_coord(grid.coord("latitude"), 0)
+        src.add_dim_coord(grid.coord("longitude"), 1)
+        self.regridder = ESMFAreaWeightedRegridder(src, tgt)
+        self.src = src
```
jamesp, line 83: Should we be timing this?

jamesp (commented):
Looks good @stephenworsley, do you have any examples of the output it produces?

Also, should we be timing the creation of the regridder object as well as using it to regrid?

benchmarks/benchmarks/ci/esmf_regridder.py
stephenworsley: I think eventually we would be timing this, but at the moment we're concerned about measuring the performance of the regridding performance because that has a bearing on using sparse vs scipy sparse .etc.

stephenworsley:
@jamesp the output is a bit hidden but you can find an example here https://cirrus-ci.com/task/4756773006475264 by clicking through the CI.

benchmarks/benchmarks/ci/esmf_regridder.py
jamesp: wouldn't the prepare stage have a dependency on the choice of sparse implementation?  It should be fairly straight forward to add an additional `time_` method here to record the performance of the init method

stephenworsley: I suppose there would be _some_ effect on prepare, but I'd expect the effect to be minimal. The vast majority of the time spent in prepare will be spent in ESMF so that any changes to the performance of initialising the sparse matrix are going to be insignificant (unless something has gone very wrong).

jamesp: If we had a benchmark we could test your expectations, and have a baseline to tell us if something has gone very wrong... 😉

stephenworsley: at any rate, I think that half of the benchmarking is probably worth its own PR.

jamesp: ok fine, please create an issue to track adding this benchmark and add the issue to the backlog.

jamesp (approved):

PR #100, opened 2021-08-04
Benchmark lazy regridding

stephenworsley:
Similar to #98 except for cubes with lazy data.

### DIFF
--- a/benchmarks/benchmarks/ci/esmf_regridder.py
+++ b/benchmarks/benchmarks/ci/esmf_regridder.py
@@ -1,12 +1,27 @@
 """Quick running benchmarks for :mod:`esmf_regrid.esmf_regridder`."""
 
+from pathlib import Path
+from shutil import rmtree
+
 import numpy as np
+import iris
 from iris.coord_systems import RotatedGeogCS
 from iris.cube import Cube
 
 from esmf_regrid.esmf_regridder import GridInfo
 from esmf_regrid.schemes import ESMFAreaWeightedRegridder
 from esmf_regrid.tests.unit.schemes.test__cube_to_GridInfo import _grid_cube
+
+
+SYNTH_DATA_DIR = Path().cwd() / "tmp_data"
+
+
+def setup_cache(*args):
+    SYNTH_DATA_DIR.mkdir(exist_ok=True)
+
+
+def teardown(*args):
+    rmtree(SYNTH_DATA_DIR)
 
 
 def _make_small_grid_args():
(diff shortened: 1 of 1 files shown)


benchmarks/benchmarks/ci/esmf_regridder.py
```diff
@@ -84,3 +88,58 @@ def setup(self, type):
 
     def time_perform_regridding(self, type):
         _ = self.regridder(self.src)
+
+
+class TimeLazyRegridding:
+    # Prevent repeat runs between setup() runs - data won't be lazy after 1st.
+    number = 1
+    # Compensate for reduced certainty by increasing number of repeats.
+    repeat = (10, 10, 10.0)
```
jamesp, line 97: Why do we have reduced certainty?  Because of the lack of `number` iterations? Is fixed 10 samples sufficient?  It looks like ASV would typically take between 2-20 with the default settings[^1].

[^1]: https://asv.readthedocs.io/en/stable/benchmarks.html#id5

benchmarks/benchmarks/ci/esmf_regridder.py
```diff
@@ -84,3 +88,58 @@ def setup(self, type):
...
+    # Prevent ASV running its warmup, which ignores `number` and would
+    # therefore get a false idea of typical run time since the data would stop
+    # being lazy.
+    warmup_time = 0.0
+
+    def setup_cache(self):
+        SYNTH_DATA_DIR = Path().cwd() / "tmp_data"
```
jamesp, line 104: I assume that `asv` deletes the current working directory at the end of the run so there is no need to clean up after ourselves here?

jamesp (commented):
Looks good, a couple of comments inline

benchmarks/benchmarks/ci/esmf_regridder.py
stephenworsley: I believe so, see this comment https://github.com/airspeed-velocity/asv/issues/511#issuecomment-300502282.

benchmarks/benchmarks/ci/esmf_regridder.py
trexfeathers: As discussed offline with @stephenworsley, below is the latest preferred setting:

```python
    # Compensate for reduced certainty by increasing number of repeats.
    #  (setup() is run between each repeat).
    #  Minimum 5 repeats, run up to 30 repeats / 20 secs whichever comes first.
    benchmark_object.repeat = (5, 30, 20.)
```

trexfeathers: ... and yes @jamesp this is due to the lack of `number` iterations, as you say.

trexfeathers (approved):

trexfeathers:
@jamesp anything left before this can be merged?

jamesp:
Seems good to me, go for it!

PR #111, opened 2021-09-06
Unstructured scheme lazy regridding (with performance tests)

stephenworsley:
Adds lazy regridding to the unstructured schemes equivalent to that added to main in #80. Also added the performance tests for unstructured schemes (including performance tests for lazy regridding) equivalent to those in #98 and #100.
Note that the main difference in code between this and #80 is the need to modify the iris function `map_complete_blocks` to be able to handle changing the number of dimensions during regridding.


esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -10,6 +11,62 @@
 from esmf_regrid.experimental.unstructured_regrid import MeshInfo
 
 
+def _map_complete_blocks(src, func, dims, out_sizes):
+    """
+    Apply a function to complete blocks.
+
+    Complete means that the data is not chunked along the chosen dimensions.
```
jamesp, line 18: Unsure what this "Complete means", means. Assuming you are referring to the sentence above, might be better to be more explicit here.

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -10,6 +11,62 @@
...
+    if not src.has_lazy_data():
+        return func(src.data)
+
+    data = src.lazy_data()
+
+    # Ensure dims are not chunked
+    in_chunks = list(data.chunks)
+    for dim in dims:
+        in_chunks[dim] = src.shape[dim]
+    data = data.rechunk(in_chunks)
```
jamesp, line 46: is this a no-op if the loop above doesn't occur?

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -10,6 +11,62 @@
...
+
+    # Determine output chunks
+    out_chunks = list(data.chunks)
+    num_dims = len(dims)
+    num_out = len(out_sizes)
+    sorted_dims = list(dims)
+    sorted_dims.sort()
```
jamesp, line 53: `sorted_dims = sorted(dims)` ?

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -10,6 +11,62 @@
...
+    dropped_dims = []
+    for i in range(max(num_dims, num_out)):
```
jamesp, line 55: This loop needs a comment - overview of what it is checking for in the multi-if.

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_GridToMeshESMFRegridder.py
```diff
@@ -196,3 +200,34 @@ def test_mask_handling():
...
+    src = Cube(src_data)
+    grid = _grid_cube(n_lons, n_lats, lon_bounds, lat_bounds, circular=True)
+    src.add_dim_coord(grid.coord("latitude"), 0)
+    src.add_dim_coord(grid.coord("longitude"), 1)
+
+    mesh_coord_x, mesh_coord_y = mesh.to_MeshCoords("face")
+    tgt_data = np.zeros([n_lats * n_lons])
+    tgt = Cube(tgt_data)
+    tgt.add_aux_coord(mesh_coord_x, 0)
+    tgt.add_aux_coord(mesh_coord_y, 0)
```
jamesp, line 226: Could any of this be refactored into a pytest fixture?

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_GridToMeshESMFRegridder.py
```diff
@@ -196,3 +200,34 @@ def test_mask_handling():
...
+
+    rg = GridToMeshESMFRegridder(src, tgt)
+
+    assert src.has_lazy_data()
+    result = rg(src)
```
jamesp, line 231: Should we be asserting the shape of the chunks / output dims?

esmf_regrid/tests/unit/experimental/unstructured_scheme/test__mesh_to_MeshInfo.py
```diff
@@ -72,6 +72,65 @@ def _example_mesh():
     return mesh
 
 
+def _gridlike_mesh(n_lons, n_lats):
```
jamesp, line 75: A small comment here for the unintiated reader would be helpful, especially the indexing / roll operations, to explain what you are creating

jamesp (changes requested):
@stephenworsley looks good, a few comments inline, mostly requests for a little more documentation.  Let's catch up on the phone and do a quick overview together too.

esmf_regrid/experimental/unstructured_scheme.py
stephenworsley: I suppose it must be, at least effectively. Also if the chunks already span the horizontal dimensions. It's worth bearing in mind that this is existing iris behaviour https://github.com/SciTools/iris/blob/d3b4c41b245ef6b4eb681f704469e991db149ba1/lib/iris/_lazy_data.py#L381

esmf_regrid/tests/unit/experimental/unstructured_scheme/test__mesh_to_MeshInfo.py
```diff
@@ -72,6 +72,86 @@ def _example_mesh():
...
+        cf_role="face_node_connectivity",
+        start_index=0,
+    )
+    lons = AuxCoord(node_lons, standard_name="longitude")
+    lats = AuxCoord(node_lats, standard_name="latitude")
+    mesh = Mesh(2, ((lons, "x"), (lats, "y")), fnc)
+
+    # In order to add a mesh to a cube, face locations must be added.
+    # These are not used in calculations and are here given a value of zero.
+    mesh_length = mesh.connectivity(contains_face=True).shape[0]
```
trexfeathers, line 147:
```suggestion
    mesh_length = mesh.face_node_connectivity.shape[0]
```

benchmarks/benchmarks/ci/esmf_regridder.py
```diff
@@ -9,8 +9,15 @@
...
 from esmf_regrid.esmf_regridder import GridInfo
+from esmf_regrid.experimental.unstructured_scheme import (
+    GridToMeshESMFRegridder,
+    MeshToGridESMFRegridder,
+)
 from esmf_regrid.schemes import ESMFAreaWeightedRegridder
 from esmf_regrid.tests.unit.schemes.test__cube_to_GridInfo import _grid_cube
+from esmf_regrid.tests.unit.experimental.unstructured_scheme.test__mesh_to_MeshInfo import (
+    _gridlike_mesh,
+)
```
trexfeathers, line 20: Please migrate these imports into just those classes/methods/functions that need them, since this will allow other benchmarks in the module to run even if the import fails.

benchmarks/benchmarks/ci/esmf_regridder.py
```diff
@@ -151,3 +158,213 @@ def time_lazy_regridding(self, cache):
...
+    # Prevent repeat runs between setup() runs - data won't be lazy after 1st.
+    number = 1
+    # Compensate for reduced certainty by increasing number of repeats.
+    #  (setup() is run between each repeat).
+    #  Minimum 5 repeats, run up to 30 repeats / 20 secs whichever comes first.
+    repeat = (5, 30, 20.0)
+    # Prevent ASV running its warmup, which ignores `number` and would
+    # therefore get a false idea of typical run time since the data would stop
+    # being lazy.
+    warmup_time = 0.0
```
trexfeathers, line 215: Put these attributes in a decorator so they can be re-used.

benchmarks/benchmarks/ci/esmf_regridder.py
```diff
@@ -151,3 +158,213 @@ def time_lazy_regridding(self, cache):
...
```
trexfeathers, line 324: Put these attributes in a decorator so they can be re-used.

benchmarks/benchmarks/ci/esmf_regridder.py
```diff
@@ -151,3 +158,213 @@ def time_lazy_regridding(self, cache):
...
+        loaded_src = iris.load_cube(file)
+        # While iris is not able to save meshes, we add these after loading.
+        mesh_coord_x, mesh_coord_y = mesh.to_MeshCoords("face")
+        loaded_src.add_aux_coord(mesh_coord_x, 0)
+        loaded_src.add_aux_coord(mesh_coord_y, 0)
+        regridder = MeshToGridESMFRegridder(loaded_src, tgt)
+
+        return regridder, file, mesh
+
+    def setup(self, cache):
```
trexfeathers, line 245: Put in a `# TODO:` to come back once Iris can save a `Mesh`, getting rid of the confusing repetition.

benchmarks/benchmarks/ci/esmf_regridder.py
```diff
@@ -151,3 +158,213 @@ def time_lazy_regridding(self, cache):
...
+        assert self.src.has_lazy_data()
+        regridder, _, _ = cache
+        _ = regridder(self.src)
+
+    def time_regridding_realisation(self, cache):
+        assert self.result.has_lazy_data()
+        _ = self.result.data
+
+
+class TimeGridToMeshRegridding:
```
trexfeathers, line 266: Could you write things in such a way that it inherits from `TimeMeshToGridRegridding` but with just a few things swapped round?

benchmarks/benchmarks/ci/esmf_regridder.py
```diff
@@ -151,3 +158,213 @@ def time_lazy_regridding(self, cache):
...
+        tgt.add_aux_coord(mesh_coord_x, 0)
+        tgt.add_aux_coord(mesh_coord_y, 0)
+        self.regridder = GridToMeshESMFRegridder(src, tgt)
+        self.src = src
+
+    def time_perform_regridding(self, type):
+        _ = self.regridder(self.src)
+
+
+class TimeLazyGridToMeshRegridding:
```
trexfeathers, line 314: Could you write things in such a way that it inherits from `TimeLazyMeshToGridRegridding` but with just a few things swapped round?

trexfeathers (changes requested):
Nice work @stephenworsley. I have quite a few comments I'm afraid! See what you think.

esmf_regrid/experimental/unstructured_scheme.py
stephenworsley: Bear in mind that this line is taken directly from iris https://github.com/SciTools/iris/blob/226bad38f91d8dddc451411900e90e7777965c75/lib/iris/_lazy_data.py#L358

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_GridToMeshESMFRegridder.py
stephenworsley: I think in general there is a lot that could be refactored for all of the tests in this directory. One of the things I've also wanted to tidy up is where helper functions (which may become fixtures) are imported from, given they are used in multiple modules. The way things are at the moment is that they exist in the test module that they were originally used in, but now we're in a situation where some helper functions are built from other helper functions that live in different modules and it could all do with a bit of tidying up. I'm unsure how we would have `test_GridToMeshESMFRegridder.py` and `test_MeshToGridESMFRegridder.py` both use the same set of fixtures without either resolving the problem  of where common helper functions come from or having one depend on the other. There may be a case for creating a separate issue for test refactoring.

One more small comment about pytest fixtures, since some helper functions are imported by the asv benchmarks, it might make sense that these remain as ordinary helper functions rather than become pytest fixtures.

trexfeathers: >some helper functions are built from other helper functions that live in different modules and it could all do with a bit of tidying up

I don't doubt it.

>There may be a case for creating a separate issue for test refactoring.

It sounds like you're saying this would be significant work in its own right? In which case I definitely agree that it should be covered later, and a task should also be raised in our internal JIRA project planner - we need to make sure this gets visibility in planning, else it is likely to remain here causing technical debt for future tasks.

trexfeathers: >since some helper functions are imported by the asv benchmarks, it might make sense that these remain as ordinary helper functions rather than become pytest fixtures.

Would writing a fixture that simply wraps the function work? From what I've read, fixtures offer a lot of benefit for test structure and clarity, especially for developers who haven't seen this conversation before looking at the tests! So it feels like it would be worth it.

benchmarks/benchmarks/ci/esmf_regridder.py
stephenworsley: I don't think these classes would work that well together, mainly because the `setup_cache` in `TimeLazyMeshToGridRegridding` additionally returns a `mesh` argument which is required in `setup`. I think this would cause confusion in the method signatures.

benchmarks/benchmarks/ci/esmf_regridder.py
```diff
@@ -151,3 +171,192 @@ def time_lazy_regridding(self, cache):
...
+        src_data = np.arange(n_lats_src * n_lons_src * h).reshape([-1, h])
+        src = Cube(src_data)
+        mesh_coord_x, mesh_coord_y = mesh.to_MeshCoords("face")
+        src.add_aux_coord(mesh_coord_x, 0)
+        src.add_aux_coord(mesh_coord_y, 0)
+        self.regridder = MeshToGridESMFRegridder(src, tgt)
+        self.src = src
+
+    def time_perform_regridding(self, type):
+        _ = self.regridder(self.src)
```
trexfeathers, line 209: You're inheriting this - doesn't need repeating.

benchmarks/benchmarks/ci/esmf_regridder.py
```diff
@@ -151,3 +171,192 @@ def time_lazy_regridding(self, cache):
...
+        tgt = Cube(tgt_data)
+        mesh = _gridlike_mesh(n_lons_tgt, n_lats_tgt)
+        mesh_coord_x, mesh_coord_y = mesh.to_MeshCoords("face")
+        tgt.add_aux_coord(mesh_coord_x, 0)
+        tgt.add_aux_coord(mesh_coord_y, 0)
+        self.regridder = GridToMeshESMFRegridder(src, tgt)
+        self.src = src
+
+    def time_perform_regridding(self, type):
+        _ = self.regridder(self.src)
```
trexfeathers, line 309: You're already inheriting this, no need to redefine.

trexfeathers (changes requested):
Thanks for the changes @stephenworsley. Just a couple of minor comments left to address.

As already agreed: please raise an issue on GitHub and on JIRA for restructuring the tests.

We discussed offline yesterday that there was no easy solution to the confusing repetition in `benchmarks/ci/esmf_regridder.py`. But I very much hope we can revisit this in future, especially once we have a better solution for synthetic data generation. I'm certain that I 90% of the lines I read appear at least twice, in different places, but in slightly different orders - there must surely be a way of generalising some things here if we take a fresh look another time!

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_GridToMeshESMFRegridder.py
stephenworsley: raised as issue #116

trexfeathers (approved):

jamesp (approved):

jamesp:
Great stuff @stephenworsley, @trexfeathers!

PR #124, opened 2021-10-26
Proposal for fixed generation of benchmark data.

trexfeathers:
Have a read of `benchmarks/benchmarks/generate_data.py` - the docstrings provide detail on my thoughts for this.

Summary: run generation scripts in an alternative environment that will therefore remain unchanged throughout the benchmark run.


benchmarks/benchmarks/generate_data.py
```diff
@@ -0,0 +1,115 @@
...
+
+        cube = original(*args, **kwargs)
+        save(cube, save_path)
+
+    save_dir = (Path(__file__).parent.parent / ".data").resolve()
+    save_dir.mkdir(exist_ok=True)
+    # TODO: caching? Currently written assuming overwrite every time.
+    save_path = save_dir / "_grid_cube.nc"
+
+    external_call = (
```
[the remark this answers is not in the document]
trexfeathers: @pp-mo this is ace! 💐 

906623c58d543427357aec48b7ca2926b2a42fa7

benchmarks/benchmarks/generate_data.py
```diff
@@ -0,0 +1,115 @@
...
+        "func("
```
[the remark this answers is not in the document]
trexfeathers: Whoops! Superseded via 906623c58d543427357aec48b7ca2926b2a42fa7

benchmarks/benchmarks/generate_data.py
```diff
@@ -0,0 +1,113 @@
...
+        Saving to a file allows the original python executable to pick back up.
+
+        Remember that all arguments must work as strings, hence the fresh
+        construction of a ``coord_system`` within the function.
+
+        """
+        from iris import save
+        from iris.coord_systems import RotatedGeogCS
+
+        from esmf_regrid.tests.unit.schemes.test__cube_to_GridInfo import _grid_cube
```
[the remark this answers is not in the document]
trexfeathers: Superseded by 906623c58d543427357aec48b7ca2926b2a42fa7

benchmarks/benchmarks/generate_data.py
```diff
@@ -35,21 +34,25 @@
     raise ValueError(error)
 
 
-def run_function_elsewhere(python_exe, func_to_run, func_call_string):
+def run_function_elsewhere(func_to_run, *args, **kwargs):
     """
-    Run a given function using an alternative python executable.
+    Run a given function using the :const:`DATA_GEN_PYTHON` executable.
```
pp-mo, line 39: I know I removed the 'python_exe' arg in my version, but it might still make more sense to retain it here.
Your choice, though, really..

trexfeathers: I flip-flopped a lot on this internally but I finally decided it was an example of me being over-flexible. I don't expect ANY examples of using more than one python executable with this function.

benchmarks/benchmarks/generate_data.py
```diff
@@ -0,0 +1,127 @@
...
+from pathlib import Path
+from textwrap import dedent
+
+from iris import load_cube
+
+#: Python executable used by :func:`run_function_elsewhere`, set via env
+#:  variable of same name. Must be path of Python within an environment that
+#:  supports iris-esmf-regrid and has iris-esmf-regrid installed via
+#:  ``pip install -e``.
+DATA_GEN_PYTHON = environ.get("DATA_GEN_PYTHON", "")
```
jamesp, line 26: Would a more sensible default value be `"python"`? otherwise suggest a normal dict lookup and have the code fail here rather than the check output line with `["","-c","a = True"]` below.

jamesp:
Take this as a suggestion, not a review. You could refactor into either a closure or a class that wraps the python executable rather than hard coding to a global constant.

e.g.

```
class PythonRunner:
  def __init__(self, python:Path):
     self.python=python
  
  def __call__(self, code, *args, **kwargs):
     ... code of run_elsewhere using self.python ...

# in the specific data gen code
run_elsewhere = PythonRunner(GEN_DATA_PYTHON)
```


or 

```
def make_python_runner(python: Path):
  def run_elsewhere(code, *args, **kwargs): ...
  return run_elsewhere

run_elsewhere = make_python_runner(GEN_DATA_PYTHON)
```

You get the picture.

benchmarks/benchmarks/generate_data.py
trexfeathers: Using the default `python` will lead to the unwanted behaviour that we are trying to avoid here - that the generated synthetic data may change with each commit.

I wanted to capture two things here - the user providing no value at all, or providing an invalid value. Perhaps this has ended up a bit hacky by me trying to encapsulate everything within one try-except. I could certainly have a catch for no value at all, and then something else for an invalid path.

trexfeathers:
>You could refactor into either a closure or a class that wraps the python executable rather than hard coding to a global constant.


If this needed to be more generic, sure. But that's engineering for something that I don't expect to happen: https://github.com/SciTools-incubator/iris-esmf-regrid/pull/124#discussion_r742927858

benchmarks/benchmarks/generate_data.py
trexfeathers: b9de1fcc580e2df822231ca637513bd9da839ca5

pp-mo (approved):
Generally between us @pp-mo @stephenworsley @jamesp we think it is OK !

PR #125, opened 2021-10-31
Add Curvilinear Support

stephenworsley:
Adds support for either the source or target grid to be curvilinear (described by 2d coordinates).

### DIFF
--- a/esmf_regrid/_esmf_sdo.py
+++ b/esmf_regrid/_esmf_sdo.py
@@ -140,9 +140,9 @@ class GridInfo(SDO):
         latdims = len(self.lats.shape)
         if len(latbounds.shape) != latdims:
             msg = (
-                f"The dimensionality of longitude bounds "
+                f"The dimensionality of latitude bounds "
                 f"({len(latbounds.shape)}) is incompatible with the "
-                f"dimensionality of the longitude ({latdims})."
+                f"dimensionality of the latitude ({latdims})."
             )
             raise ValueError(msg)
         if londims != latdims:
--- a/esmf_regrid/_esmf_sdo.py
+++ b/esmf_regrid/_esmf_sdo.py
@@ -87,8 +87,6 @@ class GridInfo(SDO):
 
     """
 
-    # TODO: Edit GridInfo so that it is able to handle 2D lat/lon arrays.
-
     def __init__(
         self,
         lons,
--- a/esmf_regrid/_esmf_sdo.py
+++ b/esmf_regrid/_esmf_sdo.py
@@ -105,16 +105,16 @@ class GridInfo(SDO):
         Parameters
         ----------
         lons : array_like
-            A 1D numpy array or list describing the longitudes of the
+            A 1D or 2D numpy array or list describing the longitudes of the
             grid points.
         lats : array_like
-            A 1D numpy array or list describing the latitudes of the
+            A 1D or 2D numpy array or list describing the latitudes of the
             grid points.
         lonbounds : array_like
-            A 1D numpy array or list describing the longitude bounds of
+            A 1D or 2D numpy array or list describing the longitude bounds of
             the grid. Should have length one greater than lons.
         latbounds : array_like
-            A 1D numpy array or list describing the latitude bounds of
+            A 1D or 2D numpy array or list describing the latitude bounds of
             the grid. Should have length one greater than lats.
         crs : cartopy projection, optional
             None or a cartopy.crs projection describing how to interpret the
@@ -130,11 +130,34 @@ class GridInfo(SDO):
         self.lons = lons
         self.lats = lats
         londims = len(self.lons.shape)
-        assert len(lonbounds.shape) == londims
+        if len(lonbounds.shape) != londims:
+            msg = (
+                f"The dimensionality of longitude bounds "
+                f"({len(lonbounds.shape)}) is incompatible with the "
+                f"dimensionality of the longitude ({londims})."
+            )
+            raise ValueError(msg)
         latdims = len(self.lats.shape)
-        assert len(latbounds.shape) == latdims
-        assert londims == latdims
-        assert londims in (1, 2)
+        if len(latbounds.shape) != latdims:
+            msg = (
+                f"The dimensionality of longitude bounds "
+                f"({len(latbounds.shape)}) is incompatible with the "
+                f"dimensionality of the longitude ({latdims})."
+            )
+            raise ValueError(msg)
+        if londims != latdims:
+            msg = (
+                f"The dimensionality of the longitude "
+                f"({londims}) is incompatible with the "
+                f"dimensionality of the latitude ({latdims})."
+            )
+            raise ValueError(msg)
+        if londims not in (1, 2):
+            msg = (
+                f"Expected a latitude/longitude with a dimensionality "
+                f"of 1 or 2, got {londims}."
+            )
+            raise ValueError(msg)
         if londims == 1:
             shape = (len(lats), len(lons))
         else:
--- a/esmf_regrid/_esmf_sdo.py
+++ b/esmf_regrid/_esmf_sdo.py
@@ -130,9 +130,9 @@ class GridInfo(SDO):
         self.lons = lons
         self.lats = lats
         londims = len(self.lons.shape)
-        assert len(self.lons.shape) == londims
+        assert len(self.lonbounds.shape) - 1 == londims
         latdims = len(self.lats.shape)
-        assert len(self.lats.shape) == latdims
+        assert len(self.latbounds.shape) - 1 == latdims
         assert londims == latdims
         assert londims in (1, 2)
         if londims == 1:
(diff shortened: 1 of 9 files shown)


esmf_regrid/_esmf_sdo.py
```diff
@@ -129,6 +129,17 @@ def __init__(
         """
         self.lons = lons
         self.lats = lats
+        londims = len(self.lons.shape)
+        assert len(self.lons.shape) == londims
```
jamesp, line 133: Are these asserts necessary? you're setting the value on the line above.

esmf_regrid/_esmf_sdo.py
```diff
@@ -129,6 +129,17 @@ def __init__(
...
+        latdims = len(self.lats.shape)
+        assert len(self.lats.shape) == latdims
+        assert londims == latdims
```
jamesp, line 136: Should we be raising Errors here? What is the preferred way of doing things in iris?  I don't usually use `assert` in library code, I would do something like:
```python
if londims != latdims:
    raise ValueError("Latitude and Longitude dimensions must be of the same order.")
```

esmf_regrid/_esmf_sdo.py
```diff
@@ -138,21 +149,37 @@ def __init__(
...
 
-        if self.circular:
-            adjustedlonbounds = self.lonbounds[:-1]
-        else:
-            adjustedlonbounds = self.lonbounds
-
-        centerlons, centerlats = np.meshgrid(self.lons, self.lats)
-        cornerlons, cornerlats = np.meshgrid(adjustedlonbounds, self.latbounds)
+        londims = len(self.lons.shape)
+        assert len(self.lons.shape) == londims
```
jamesp, line 161: Why are we repeating all the shape checking? Is there a case where it is not caught in `__init__`? i.e. are `self.lons` and `self.lats` assigned to / changed elsewhere?

jamesp: If it's necessary, should probably be pulled out into e.g. `self._check_dim_shape()`.

esmf_regrid/_esmf_sdo.py
```diff
@@ -138,21 +149,37 @@ def __init__(
...
+        if londims == 1:
+            if self.circular:
+                adjustedlonbounds = self.lonbounds[:-1]
+            else:
+                adjustedlonbounds = self.lonbounds
+            centerlons, centerlats = np.meshgrid(self.lons, self.lats)
+            cornerlons, cornerlats = np.meshgrid(adjustedlonbounds, self.latbounds)
+        elif londims == 2:
+            if self.circular:
+                slice = np.s_[:, :-1]
```
jamesp, line 176: coverage says that this line is not being hit by the tests. so I assume that means we don't have a test case for circular curvilinear data. should we?

esmf_regrid/schemes.py
```diff
@@ -4,9 +4,9 @@
 import copy
 import functools
 
+from cf_units import Unit
```
jamesp, line 7: Is this the first time we're importing cf_units? It needs to be added to the requirements.yml files if so.  I expect we get it as a transitive dep from iris, but good to be explicit.

esmf_regrid/schemes.py
```diff
@@ -18,39 +18,43 @@
...
-    assert isinstance(lat, iris.coords.DimCoord)
-    # TODO: accommodate other x/y coords.
-    # TODO: perform checks on lat/lon.
+    lon, lat = cube.coord(axis="x"), cube.coord(axis="y")
     #  Checks may cover units, coord systems (e.g. rotated pole), contiguous bounds.
     if cube.coord_system() is None:
         crs = None
     else:
         crs = cube.coord_system().as_cartopy_crs()
+    londim, latdim = len(lon.points.shape), len(lat.points.shape)
```
jamesp, line 33: This is operating on the `points` property, above when doing these checks we were operating on the object itself. Is this because we now have iris objects instead of numpy arrays?

esmf_regrid/schemes.py
```diff
@@ -18,39 +18,43 @@
...
+        # Ensure coords come from a proper grid.
+        assert isinstance(lon, iris.coords.DimCoord)
+        assert isinstance(lat, iris.coords.DimCoord)
+        circular = lon.circular
+        # TODO: perform checks on lat/lon.
+    elif londim == 2:
+        assert cube.coord_dims(lon) == cube.coord_dims(lat)
+        assert lon.is_contiguous()
+        assert lat.is_contiguous()
+        circular = False
```
jamesp, line 46: Why is this always false? Above in `GridInfo` we have support for circular 2d coords. What is different here? A comment explaining would be useful.

esmf_regrid/schemes.py
```diff
@@ -121,8 +125,15 @@ def _create_cube(data, src_cube, grid_dim_x, grid_dim_y, grid_x, grid_y):
     """
     new_cube = iris.cube.Cube(data)
 
-    new_cube.add_dim_coord(grid_x, grid_dim_x)
-    new_cube.add_dim_coord(grid_y, grid_dim_y)
+    if len(grid_x.points.shape) == 1:
+        new_cube.add_dim_coord(grid_x, grid_dim_x)
+    else:
+        new_cube.add_aux_coord(grid_x, (grid_dim_y, grid_dim_x))
```
jamesp, line 131: Is it sufficient to only add an AuxCoord or do we need to add a dim coord too?

esmf_regrid/schemes.py
```diff
@@ -147,11 +158,16 @@ def copy_coords(src_coords, add_method):
...
-    src_x, src_y = get_xy_dim_coords(src_grid_cube)
-
-    grid_x_dim = src_grid_cube.coord_dims(src_x)[0]
-    grid_y_dim = src_grid_cube.coord_dims(src_y)[0]
+    tgt_x = tgt_grid_cube.coord(axis="x")
+    tgt_y = tgt_grid_cube.coord(axis="y")
+    src_x = src_grid_cube.coord(axis="x")
+    src_y = src_grid_cube.coord(axis="y")
+
+    if len(src_x.points.shape) == 1:
```
jamesp, line 166: As I read through this, I wonder if a utility function might make things more readable:

```python
if not is_rectilinear(src_x):  # instead of if len(src_x.points.shape) == 1:
   ...
```
where the util function is something like:
```python
def is_rectilinear(coord) -> bool:
    if hasattr(coord, 'points'):
        shape = coord.points.shape
    else:
        shape = coord.shape
    return len(shape) == 2
````

jamesp (changes requested):
Looks good @stephenworsley some comments inline.

There is a comment at L90 of _esmf_sdo.py, can that be removed now?
Also I think the signatures of some functions in the docstrings need updating as they reference 1d objects, but if I understand your code correctly that may no longer be true! 🎉 

And finally a more general point, it might be useful to start thinking about documentation for this - either as comments in the code or in an rst file somewhere. What can it do, what can't it do, perhaps a table of compatibility?!

esmf_regrid/_esmf_sdo.py
stephenworsley: ah, good catch, I believe I may have intended this assert to be for lonbounds.

esmf_regrid/_esmf_sdo.py
stephenworsley: Yes, I think i have been using assert as a bit of a shortcut. I'll see what I can do about raising proper errors.

esmf_regrid/_esmf_sdo.py
stephenworsley: I think you're right that this is redundant. this information shouldn't be changing from initialisation to here.

esmf_regrid/_esmf_sdo.py
stephenworsley: So it turns out that only `DimCoord`s have the `circular` property, so this is not going to be hit via the iris scheme. With that said, it is theoretically possible to call this by setting up a regridder directly. In that sense the code here is complete with respect to the core of esmf-regrid and the testing is complete with respect to the iris schemes but there is now a hole in the testing with respect to the core of esmf-regrid. I suppose it's probably worth adding a test for this case for completeness sake.

esmf_regrid/schemes.py
stephenworsley: I believe thats correct, here `lon` represents a coord whereas in `_esmf_sdo`, `lons` represented an array.

esmf_regrid/schemes.py
stephenworsley: I don't think there is a DimCoord we could add here.

jamesp:
Looks good to go 👍 

Things to do after this is merged:
- [ ] Bump version number to 0.3.0 and tag as release
- [ ] Add 0.3.0 changes to the changelog
- [ ] Bump version number to 0.4.dev0.
- [ ] schemes.py#L36 needs an issue to track the additional checks required. Add to backlog.

jamesp (approved):

PR #128, opened 2021-11-05
Update dependencies

trexfeathers:
Includes coverage packages in project dependencies and updates the lockfiles.


trexfeathers:
@stephenworsley I'd appreciate your thoughts on why the changed dependencies have caused regressions. I've tried re-running the benchmarking job but they're still there.

stephenworsley:
For `time_make_grid`, my guess would be something to do with cartopy running these lines:
https://github.com/SciTools-incubator/iris-esmf-regrid/blob/82c494875894b7e54fd8c6c2f6707b917d26eac2/esmf_regrid/_esmf_sdo.py#L157-L158
Cartopy has gone up a version here, as has Proj. I'm not sure about the other slowdowns.

trexfeathers:
>I'm not sure about the other slowdowns.

I don't suppose they might also depend on the same stuff as `time_make_grid`? Which, as you say, we already understand (and have merged on `main`).

stephenworsley (approved):
Looks good, the major benchmark slowdowns seem to be due to cartopy and are understood.

PR #130, opened 2021-11-09
Regridder load/saving

stephenworsley:
Adds a save/load function for the unstructured regridders. Also, allows initialisation of regridders from precomputed weights.


esmf_regrid/experimental/io.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # Determine the regridder type.
+    regridder_type = metadata_cube.attributes["regridder type"]
+    if regridder_type == "grid to mesh":
+        scheme = GridToMeshESMFRegridder
+    elif regridder_type == "mesh to grid":
+        scheme = MeshToGridESMFRegridder
+    else:
+        raise ValueError(
+            f"expected a regridder type to be either 'grid to mesh' "
```
bjlittle, line 130: @stephenworsley You can turn `codecov` patch warnings off with `patch: off` in the `codecov.yml`, see https://github.com/SciTools/cf-units/blob/main/codecov.yml

I found `codecov` patching an annoyance, particularly as it causes CI to fail

esmf_regrid/experimental/io.py
```diff
@@ -0,0 +1,150 @@
...
+        The regridder instance to save.
+    filename : str
+        The file name to save to.
+    """
+    src_name = "regridder source field"
+    tgt_name = "regridder target field"
+    if isinstance(rg, GridToMeshESMFRegridder):
+        regridder_type = "grid to mesh"
+        src_grid = (rg.grid_y, rg.grid_x)
+        src_shape = [len(coord.points) for coord in src_grid]
```
bjlittle, line 34: Shame you can't do `src_grid.shape`... or can you?

Might be a handy property to have, as it'll avoid such boiler-plate code as above.

(not necessary to fix for this PR btw - just an observation 👍)

esmf_regrid/experimental/io.py
```diff
@@ -0,0 +1,150 @@
...
+    ----------
+    rg : GridToMeshESMFRegridder, MeshToGridESMFRegridder
+        The regridder instance to save.
+    filename : str
+        The file name to save to.
+    """
+    src_name = "regridder source field"
+    tgt_name = "regridder target field"
+    if isinstance(rg, GridToMeshESMFRegridder):
+        regridder_type = "grid to mesh"
```
bjlittle, line 32: Could you not just use `rg.__class__.__name__` instead here?

That way it scales to supporting other regridders, and explicitly specific about the regridder under the hood.

Don't know if this matters... or whether you want the `regridder_type` just to be generic rather than specific to a class 🤔

esmf_regrid/tests/__init__.py
```diff
@@ -69,3 +72,48 @@ def make_grid_args(nx, ny):
...
+        Optional filename extension.
+
+    Returns
+    -------
+    str
+        An appropriate filename for a temporary file.
+    """
+    temp_file = tempfile.mkstemp(suffix)
+    os.close(temp_file[0])
+    return temp_file[1]
```
bjlittle, line 95: @stephenworsley You're living in the wonderful world of `pytest` and there are super handy `pytest` fixtures that do this for you for free... see https://docs.pytest.org/en/6.2.x/tmpdir.html 😉 

This would remove the need for both `create_temp_filename` and `temp_filename` 👍

bjlittle (changes requested):
@stephenworsley Lovely stuff 😄 

Just a couple of comment to service and we're good to go AFAIC

esmf_regrid/tests/__init__.py
stephenworsley: I was trying to immitate iris functions here, it makes sense that pytest would be able to clear things up a bit, I'll have a look at these.

esmf_regrid/experimental/io.py
stephenworsley: Well at this point I think `src_grid` is just a tuple of coords so I don't believe it would have a `.shape`.

esmf_regrid/experimental/io.py
stephenworsley: I'll change this for the time being because it tidies up some of the code, though this question does seem worth considering for when we are refining this.

esmf_regrid/tests/unit/experimental/io/test_round_tripping.py
```diff
@@ -0,0 +1,118 @@
...
+    src_data = np.zeros(src_lons * src_lats)
+    src = Cube(src_data)
+    src.add_aux_coord(mesh_coord_x, 0)
+    src.add_aux_coord(mesh_coord_y, 0)
+
+    rg = MeshToGridESMFRegridder(src, tgt, mdtol=0.5)
+    return rg, src
+
+
+def test_GridToMeshESMFRegridder_round_trip(tmp_path):
```
bjlittle, line 64: @stephenworsley Lovely stuff :man_singer:

bjlittle (approved):

PR #138, opened 2021-12-10
Check mesh equality on MeshToGridESMFRegridder call

stephenworsley:
Now that iris provides proper mesh equality comparison, this can be done during the MeshToGridESMFRegridder call.

### DIFF
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -432,22 +432,25 @@ class MeshToGridESMFRegridder:
 
         """
         mesh = cube.mesh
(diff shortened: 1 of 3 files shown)


stephenworsley:
Current errors are due to saved and reloaded regridders having a mesh which does not equate to the original mesh which initialised the regridder. I have tried to fix this by equalising the `var_name`s attached to the mesh, but there still appears to be an error due to masked arrays in connectivities being filled inconsistently on equality checks.
When this line:
https://github.com/SciTools/iris/blob/68b5e65e4cca34505f4731d2e9f3794cbd1d4f16/lib/iris/util.py#L366
is changed to `array = ma.asarray(array)` thsi appears to fix the problem.

PR #144, opened 2022-01-26
Use more stable version of connectivity indices

stephenworsley:
Using `indices_by_location()` ensures that regridding still works in the case where `face_node_connectivity.location_axis` = 1.


trexfeathers:
Also need a change here:

https://github.com/stephenworsley/iris-esmf-regrid/blob/9af2e801eb1af1ec598790e102ce5fef3a49a6a8/esmf_regrid/experimental/unstructured_scheme.py#L583

noxfile.py
```diff
@@ -16,13 +16,13 @@
 
 
 #: Default to reusing any pre-existing nox environments.
-nox.options.reuse_existing_virtualenvs = True
+nox.options.reuse_existing_virtualenvs = False
```
trexfeathers, line 19: Do we want to keep this? Will it break this PR's CI to change it back to `True`?

trexfeathers (approved):
I'm merging including the benchmark failure, which we've determined is statistical noise (need to think more about sensitivity).

PR #147, opened 2022-02-01
Update feature branch

stephenworsley:
Merge feature branch into a copy of main.


trexfeathers:
This has all the commits I would expect, and no others. But from previous updates (both here and on Iris feature branches) I would have expected a commit that looks like 9384c99feea2fefdaee8ca7bed1d59ed91c52e40:

>Updating feature branch unstructured_scheme from 44d6048 to head of main, e528cbf

Any ideas?

stephenworsley:
Good catch @trexfeathers, it looks like I missed the step where you ammend the commit message.

PR #148, opened 2022-02-10
Add bilinear regridding support

stephenworsley:
Adds support for bilinear regridding for the unstructured regridders.


trexfeathers:
@stephenworsley #162 might solve the CI failures (unless you think the regressions are genuine)?

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -476,7 +539,7 @@ def _regrid_along_grid_dims(regridder, data, grid_x_dim, grid_y_dim, mdtol):
     return result
 
 
-def _create_mesh_cube(data, src_cube, grid_x_dim, grid_y_dim, mesh):
+def _create_mesh_cube(data, src_cube, grid_x_dim, grid_y_dim, mesh, location="face"):
```
trexfeathers, line 542:
```suggestion
def _create_mesh_cube(data, src_cube, grid_x_dim, grid_y_dim, mesh, location):
```

I've been through the call stack. This is always going to receive a `location`.

(Indeed if it doesn't receive a `location` that's probably breaking some assumptions about how the code is working and we want an `Error` rather than silent continuation).

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -501,6 +564,8 @@ def _create_mesh_cube(data, src_cube, grid_x_dim, grid_y_dim, mesh):
     mesh : Mesh
         The :class:`iris.experimental.ugrid.Mesh` for the new
         Cube.
+    location : str, default="face"
```
trexfeathers, line 567:
```suggestion
    location : str
```

See above.

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -536,7 +601,10 @@ def copy_coords(src_coords, add_method):
 
 
 def _regrid_rectilinear_to_unstructured__prepare(
-    src_grid_cube, target_mesh_cube, precomputed_weights=None
+    src_grid_cube,
+    target_mesh_cube,
+    method="conservative",
```
trexfeathers, line 606:
```suggestion
    method,
```

Only called by `GridToMeshESMFRegridder` and `regrid_rectilinear_to_unstructured`, both of which have the default covered - those being the 'public ways in'.

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -547,18 +615,37 @@ def _regrid_rectilinear_to_unstructured__prepare(
...
+        center = False
+    elif method == "bilinear":
+        if location not in ["face", "node"]:
+            raise ValueError(
+                f"Bilinear regridding requires a target cube with a node "
+                f"or face location, target cube had the {location} location."
+            )
+        center = True
+    else:
+        raise ValueError("method must be either bilinear or conservative.")
```
trexfeathers, line 636: Use a `NotImplementedError`, referencing the unsupported argument that was provided.

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -644,14 +736,19 @@ def regrid_rectilinear_to_unstructured(src_cube, mesh_cube, mdtol=0):
         target cell. ``mdtol=0`` means no missing data is tolerated while ``mdtol=1``
         will mean the resulting element will be masked if and only if all the
         overlapping cells of the ``src_cube`` are masked.
+    method : str, default="conservative"
+        The regridding method which :mod:`ESMF` uses to calculate weights. Can be
+        either "conservative" or "bilinear".
```
trexfeathers, line 741: Could reference the original class attributes, to make this less of a closed box, e.g. 

```
:attr:`ESMF.RegridMethod.CONSERVE`
```

Or is that overkill?

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -660,7 +757,12 @@ class GridToMeshESMFRegridder:
     r"""Regridder class for rectilinear to unstructured :class:`~iris.cube.Cube`\\ s."""
 
     def __init__(
-        self, src_mesh_cube, target_grid_cube, mdtol=1, precomputed_weights=None
+        self,
+        src_mesh_cube,
+        target_grid_cube,
+        mdtol=None,
```
trexfeathers, line 763: Also need to change the docstring to:

```
mdtol : float, optional
```

... with possible references to the defaults for the different `method` values.

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -389,26 +431,41 @@ def __init__(
             exceeds ``mdtol``. ``mdtol=0`` means no missing data is tolerated while
             ``mdtol=1`` will mean the resulting element will be masked if and only
             if all the contributing elements of data are masked.
+        method : str, default="conservative"
+            The regridding method which :mod:`ESMF` uses to calculate weights. Can be
+            either "conservative" or "bilinear".
```
trexfeathers, line 436: Could reference the original class attributes, to make this less of a closed box, e.g. 

```
:attr:`ESMF.RegridMethod.CONSERVE`
```

Or is that overkill?

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -677,21 +779,43 @@ def __init__(
...
```
trexfeathers, line 784: Could reference the original class attributes, to make this less of a closed box, e.g. 

```
:attr:`ESMF.RegridMethod.CONSERVE`
```

Or is that overkill?

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -677,21 +779,43 @@ def __init__(
...
 
         """
+        if method not in ["conservative", "bilinear"]:
+            raise (ValueError("method must be either bilinear or conservative."))
```
trexfeathers, line 788: Use a `NotImplementedError`, referencing the invalid argument that was provided.

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -677,21 +779,43 @@ def __init__(
...
             msg = "Value for mdtol must be in range 0 - 1, got {}."
             raise ValueError(msg.format(mdtol))
         self.mdtol = mdtol
+        self.method = method
 
         partial_regrid_info = _regrid_rectilinear_to_unstructured__prepare(
-            src_mesh_cube, target_grid_cube, precomputed_weights=precomputed_weights
+            src_mesh_cube,
+            target_grid_cube,
+            method=method,
```
trexfeathers, line 805:
```suggestion
            method=self.method,
```

Less risk of future divergence.

trexfeathers (changes requested):
OK @stephenworsley, here's my review of the functionality.

Please could you also merge in the latest `main`, which should avoid the various benchmarking failures we seem to be suffering.

I will move on to reviewing the tests next.

esmf_regrid/_esmf_sdo.py
```diff
@@ -127,6 +128,9 @@ def __init__(
             Array describing the areas associated with
             each face. If ``None``, then :mod:`ESMF` will use its own
             calculated areas.
+        center : bool, default=False
```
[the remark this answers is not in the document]
stephenworsley: ESMF uses `center` so I think I'll stick with that. I'm hesitant to lean too much into UK english when writing variable names, especially for common fairly words like "color" that get used in lots of different packages.

esmf_regrid/esmf_regridder.py
```diff
@@ -79,11 +82,21 @@ def __init__(self, src, tgt, precomputed_weights=None):
         self.src = src
         self.tgt = tgt
 
+        if method == "conservative":
+            esmf_regrid_method = ESMF.RegridMethod.CONSERVE
+        elif method == "bilinear":
+            esmf_regrid_method = ESMF.RegridMethod.BILINEAR
+        else:
+            raise ValueError("method must be either bilinear or conservative.")
```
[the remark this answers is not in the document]
stephenworsley: While i think it's a good idea to reference the invalid argument, I'm not sure `NotImplementedError` is the appropriate error to use here. Not all values which will raise this error will be things we are likely to ever implement, so i feel like the more generic `ValueError` works better here. I could see `NotImplementedError` being used as well to catch specific methods we had a view to include some day, though I feel like that may be overkill at this point and I don't think there are any methods I want to commit to at this point.

esmf_regrid/esmf_regridder.py
```diff
@@ -79,11 +82,21 @@ def __init__(self, src, tgt, precomputed_weights=None):
...
+        self.method = method
+
         self.esmf_regrid_version = esmf_regrid.__version__
         if precomputed_weights is None:
             self.esmf_version = ESMF.__version__
             weights_dict = _get_regrid_weights_dict(
-                src.make_esmf_field(), tgt.make_esmf_field()
+                src.make_esmf_field(),
+                tgt.make_esmf_field(),
+                regrid_method=esmf_regrid_method,
```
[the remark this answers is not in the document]
stephenworsley: I don't think that works here, `esmf_regrid_method` is an ESMF object while `self.method` is a string. The reason we hold `self.method` as a string is because it's used in saving/loading.

esmf_regrid/experimental/io.py
```diff
@@ -57,15 +58,17 @@ def save_regridder(rg, filename):
         src_cube.add_dim_coord(src_grid[1], 1)
 
         tgt_mesh = rg.mesh
-        tgt_data = np.zeros(tgt_mesh.face_node_connectivity.indices.shape[0])
+        tgt_location = rg.location
+        tgt_data = np.zeros(tgt_mesh.to_MeshCoords(tgt_location)[0].points.shape[0])
```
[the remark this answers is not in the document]
stephenworsley: Do let my know if you find an easier way to write this.

esmf_regrid/experimental/unstructured_regrid.py
```diff
@@ -35,7 +37,7 @@ def __init__(
         node_coords: :obj:`~numpy.typing.ArrayLike`
             An ``Nx2`` array describing the location of the nodes of the mesh.
             ``node_coords[:,0]`` describes the longitudes in degrees and
-            ``node_coords[:,1]`` describes the latitudes in degrees
+            ``node_coords[:,1]`` describes the latitudes in degrees.
         face_node_connectivity: :obj:`~numpy.typing.ArrayLike`
```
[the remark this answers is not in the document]
stephenworsley: in theory, this could become optional, but at the moment MeshInfo is still built around 2D meshes. I think the case of bilinear regridding on non 2D meshes is probably too much of an edge case at the moment to be worth the effort here.

esmf_regrid/esmf_regridder.py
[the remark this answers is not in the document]
trexfeathers: Oops! Cheerfully withdrawn

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -116,18 +116,24 @@ def _bounds_cf_to_simple_1d(cf_bounds):
...
 
 
-def _mesh_to_MeshInfo(mesh):
+def _mesh_to_MeshInfo(mesh, location="face"):
     # Returns a MeshInfo object describing the mesh of the cube.
     assert mesh.topology_dimension == 2
+    if None in mesh.face_coords:
+        elem_coords = None
+    else:
+        elem_coords = np.stack([coord.points for coord in mesh.face_coords], axis=-1)
```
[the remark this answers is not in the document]
trexfeathers: Yes. Either 1 or 0 `face_coords`

trexfeathers (changes requested):
I've reviewed the tests somewhat:

* Read the coverage report to confirm that the additional code is being exercised in the tests.
* Confirmed that the tests are passing.
* Skim read the tests themselves to check that they make sense.
  (Not being an expert in this area, I can't pretend to have been super-thorough).

I think that's enough for what we need!

Couple of requests for you.

esmf_regrid/tests/unit/experimental/unstuctured_regrid/test_MeshInfo.py
```diff
@@ -85,3 +85,75 @@ def _give_extra_dims(array):
...
+    #      0  10
+    grid_args = [ar * 2 for ar in make_grid_args(2, 3)]
+    grid = GridInfo(*grid_args, center=True)
+
+    mesh_to_grid_regridder = Regridder(node_mesh, grid, method="bilinear")
+    mesh_input = np.arange(5)
+    grid_output = mesh_to_grid_regridder.regrid(mesh_input)
+    # For a flat surface, we would expect the fractional part of these values
+    # to be either 1/3 or 2/3. Since the actual surface lies on a sphere, and
+    # due to the way ESMF approximates these values, the expected output is
```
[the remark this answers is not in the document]
stephenworsley: Perhaps "calculates" is a better word than "approximates". I guess my point is that the values will be close to our intuition for flat surfaces but will necessarily be different since we are not flat. You could view this as a way of approximating the behaviour of flat bilinear regridding, but I don't think I can say that ESMF is _wrong_ here, since I'm not entirely sure what the correct behaviour _ought_ to be in this case. Since the numbers are close enough to my intuition, I'm giving ESMF the benefit of the doubt that the "approximation" here is an appropriate one.

esmf_regrid/esmf_regridder.py
```diff
@@ -71,6 +71,9 @@ def __init__(self, src, tgt, precomputed_weights=None):
             Describes the target mesh/grid.
             Data output by this regridder will be a :class:`numpy.ndarray` whose
             shape is compatible with ``tgt``.
+        method : str
+            Either "conservative" or "bilinear". Describes the regrid
+            method :mod:`ESMF` uses to calculate weights.
```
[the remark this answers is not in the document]
trexfeathers: Just tested this and the correct path is actually

```
:attr:`ESMF.api.constants.RegridMethod.CONSERVE`
```

So maybe `~ESMF...` is called for.

Sorry for missing that!

esmf_regrid/experimental/unstructured_scheme.py
trexfeathers: Just tested this and the correct path is actually

```
:attr:`ESMF.api.constants.RegridMethod.CONSERVE`
```

So maybe `~ESMF...` is called for.

Sorry for missing that!

esmf_regrid/experimental/unstructured_scheme.py
trexfeathers: Just tested this and the correct path is actually

```
:attr:`ESMF.api.constants.RegridMethod.CONSERVE`
```

So maybe `~ESMF...` is called for.

Sorry for missing that!

trexfeathers (approved):
Great job @stephenworsley, looking forward to our users enjoying this new feature 🙂

PR #152, opened 2022-02-22
Add brief overview to readme

stephenworsley:
Add enough documentation to the readme so that new users should be able to make use of the most important features by following the example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,3 +7,37 @@
 [![License](https://img.shields.io/github/license/SciTools-incubator/iris-esmf-regrid)](https://github.com/SciTools-incubator/iris-esmf-regrid/blob/main/LICENSE)
 [![Contributors](https://img.shields.io/github/contributors/SciTools-incubator/iris-esmf-regrid)](https://github.com/SciTools-incubator/iris-esmf-regrid/graphs/contributors)
 ![Mark stale issues and pull requests](https://github.com/SciTools-incubator/iris-esmf-regrid/workflows/Mark%20stale%20issues%20and%20pull%20requests/badge.svg)
+
+---
+
+## Overview
+
+This project aims to provide a bridge between [iris](https://github.com/SciTools/iris)
+and [ESMF](https://github.com/esmf-org/esmf). This takes the form of regridder classes
+which take iris cubes as their arguments and use ESMF to perform regridding
+calculations. These classes are designed to perform well on cubes which have multiple
+non-horizontal dimensions and lazy ([dask](https://github.com/dask/dask)) data.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,3 +7,60 @@
...
+# onto a cube with the same grid as `target_grid_cube`.
+result = regridder(source_mesh_cube)
+```
+
+Note that this pattern allows the reuse of an initialised regridder so long as the
+source and target grid/mesh are the same. Since initialisation often contains the
+most computationally intensive part of the regridding process, this pattern can save
+significant amounts of time when regridding. To make use of this efficiency across
+sessions, we support the saving of certain regridders (currently
+`MeshToGridESMFRegridder` and `GridToMeshESMFRegridder`). We can do this as follows:
```
trexfeathers, line 53: Don't bother listing which ones, you'll only have to change later.

README.md
```diff
@@ -7,3 +7,60 @@
...
+
```
trexfeathers, line 54: I feel like we could get the same point across with half the characters. This is the first thing people see, let's make it snappy 😁

trexfeathers (changes requested):
Thanks @stephenworsley, this does a good job of getting the point across. But we can make it shorter/easier; that's what I've seen when looking at other repos that rely on their `README` as the ice-breaker:

* https://github.com/actions/cache
* https://github.com/SciTools/cf-units

trexfeathers:
Already looking loads better IMO 🙂

zklaus:
I think the main path to regridding should still follow the standard approach in Iris for regridding as documented [here](https://scitools-iris.readthedocs.io/en/latest/userguide/interpolation_and_regridding.html#regridding), don't you think?

As such, the first and principal example should be something along the lines of
```
result = source_cube.regrid(target_cube, esmf_regrid.ESMFRegrid())
```

stephenworsley:
@zklaus Ultimately, I think the documentation probably would benefit from looking more like the iris documentation. With that said, the current implementation of regridders does not go so far as to offer an exact match to iris functionality. This documentation reflects that and describes the regridding functionality which is common between the structured and unstructured regridders. The cached regridder is also necessary to introduce to discus regridder saving, which is a new feature not contained in iris. The unstructured and structured regridders are currently a little disjointed in their functionality and I hope to be unify them in a future release, however now that UGRID support in iris has released, I'm keen to release what we have for unstructured regridding as soon as possible, even if the structured and unstructured regridders do not yet have the same level of functionality.

trexfeathers (approved):

PR #155, opened 2022-02-23
Automatic API Documentation

trexfeathers:
* [x] Create Sphinx configuration
* [x] Create Read the Docs configuration
* [x] Populate docstrings with Sphinx domains

[Here is a rendered example on Read the Docs](https://trexfeathers-iris-esmf-regrid.readthedocs.io/en/latest/)


LICENSE
```diff
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2020, SciTools-incubator
+Copyright (c) 2020 - 2022, SciTools-incubator
```
bjlittle, line 3: Not sure that we need to add an end date to this, some projects do and some don't

I'm not a lawyer, but don't see the harm :+1:

trexfeathers:
[Here is a rendered example on Read the Docs](https://trexfeathers-iris-esmf-regrid.readthedocs.io/en/latest/)

esmf_regrid/schemes.py
```diff
@@ -225,33 +225,35 @@ def _regrid_rectilinear_to_rectilinear__perform(src_cube, regrid_info, mdtol):
...
+    values calculated using the area weighted
+    mean of :attr:`~iris.cube.Cube.data` values from ``src_cube`` regridded onto the
+    horizontal grid of ``grid_cube``.
 
     Parameters
     ----------
-    src_cube : cube
-        An urectilinear instance of iris.cube.Cube that supplies the data,
+    src_cube : :class:`iris.cube.Cube`
+        An un-rectilinear instance of :class:`~iris.cube.Cube` that supplies the data,
```
stephenworsley, line 238: I believe this should have been "rectilinear". With that said, I have a feeling we also support curvilinear grids now. This could be updated in a later PR if needs be since "rectilinear" is in the name of the function and I can't think up a good name to replace it that covers both rectilinear and curvilinear grids.

LICENSE
bjlittle: I've no strong feelings either way. Totally happy to leave it as you've changed it.

Perhaps raise it as a post discussion with other core devs? That seems reasonable to me :+1:

bjlittle (changes requested):
@trexfeathers Awesome LGTM.

Just a minor question to service, but let's get this banked :100: :man_singer:

esmf_regrid/schemes.py
trexfeathers: 4a82f52691364fb89498abaa42867a46ca0e63cf

stephenworsley (approved):

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -689,21 +695,23 @@ def __init__(
...
+            A :class:`~iris.cube.Cube` instance to be regridded.
 
         Returns
         -------
-            A cube defined with the horizontal dimensions of the target
-            and the other dimensions from this cube. The data values of
-            this cube will be converted to values on the new grid using
+        :class:`iris.cube.Cube`
+            A :class:`~iris.cube.Cube` defined with the horizontal dimensions of the target
+            and the other dimensions from this :class:`~iris.cube.Cube`. The
```
stephenworsley, line 712: I think this should refer to the argument "cube" rather than the class "Cube".

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -413,21 +415,23 @@ def __init__(
...
```
stephenworsley, line 432: I think this should refer to the argument "cube" rather than the class "Cube".

stephenworsley (changes requested):
Looking good, just some minor tweaks and I think we're good.

esmf_regrid/experimental/unstructured_scheme.py
trexfeathers: 99dd4241559692dd5174d53618288c0c5af1c514

esmf_regrid/experimental/unstructured_scheme.py
trexfeathers: 99dd4241559692dd5174d53618288c0c5af1c514

stephenworsley (approved):
Looks good!

PR #162, opened 2022-03-09
Increase standard benchmark regression tolerance

trexfeathers:
Benchmarks are currently over-sensitive and regularly flag false positive regressions.


stephenworsley:
I suspect that we might also want to reduce the number of test cases we use as well for the sake of time, though given the kind of variability we're seeing this is probably also necessary.

stephenworsley (approved):
Looks good.

PR #175, opened 2022-04-27
Unify unstructured and rectilinear schemes (curvilinear support)

stephenworsley:
Generalises the helper functions (`_cube_to_GridInfo` and `_create_cube`) so that one of each type of function can be used for all regridding schemes. This reduces code copying and begins to unify the behaviour of the regridders.

This adds curvilinear support to the unstructured regridders.

### DIFF
--- a/esmf_regrid/schemes.py
+++ b/esmf_regrid/schemes.py
@@ -6,6 +6,7 @@ import functools
 
 from cf_units import Unit
 import iris
+from iris.coords import DimCoord
 from iris._lazy_data import map_complete_blocks
 import numpy as np
 
--- a/esmf_regrid/schemes.py
+++ b/esmf_regrid/schemes.py
@@ -165,8 +165,7 @@ def _create_cube(data, src_cube, src_dims, tgt_coords, num_tgt_dims):
                 if src_dim in dims:
                     continue
             offset = num_tgt_dims - len(src_dims)
-            if len(src_dims) == 1:
-                dims = [dim if dim < max(src_dims) else dim + offset for dim in dims]
+            dims = [dim if dim < max(src_dims) else dim + offset for dim in dims]
             result_coord = coord.copy()
             # Add result_coord to the owner of add_method.
             add_method(result_coord, dims)
--- a/esmf_regrid/schemes.py
+++ b/esmf_regrid/schemes.py
@@ -58,7 +58,6 @@ def _cube_to_GridInfo(cube, center=False, resolution=None):
             crs=crs,
             circular=circular,
             center=center,
-            resolution=resolution,
         )
     else:
         grid_info = RefinedGridInfo(
--- a/esmf_regrid/schemes.py
+++ b/esmf_regrid/schemes.py
@@ -146,7 +146,7 @@ def _create_cube(data, src_cube, src_dims, tgt_coords, num_tgt_dims):
         )
     if num_tgt_dims == 1:
         grid_dim_x = grid_dim_y = min(src_dims)
-    for tgt_coord, dim in tgt_coords, (grid_dim_x, grid_dim_y):
+    for tgt_coord, dim in zip(tgt_coords, (grid_dim_x, grid_dim_y)):
         if len(tgt_coord.shape) == 1:
             if isinstance(tgt_coord, DimCoord):
                 new_cube.add_dim_coord(tgt_coord, dim)
--- a/esmf_regrid/schemes.py
+++ b/esmf_regrid/schemes.py
@@ -18,7 +18,7 @@ __all__ = [
 ]
 
 
-def _cube_to_GridInfo(cube):
+def _cube_to_GridInfo(cube, center=False, resolution=None):
     # This is a simplified version of an equivalent function/method in PR #26.
     # It is anticipated that this function will be replaced by the one in PR #26.
     #
@@ -49,14 +49,24 @@ def _cube_to_GridInfo(cube):
     lon_bound_array = lon.units.convert(lon_bound_array, Unit("degrees"))
     lat_bound_array = lat.contiguous_bounds()
     lat_bound_array = lat.units.convert(lat_bound_array, Unit("degrees"))
-    return GridInfo(
-        lon.points,
-        lat.points,
-        lon_bound_array,
-        lat_bound_array,
-        crs=crs,
-        circular=circular,
-    )
+    if resolution is None:
+        grid_info = GridInfo(
+            lon.points,
+            lat.points,
+            lon_bound_array,
+            lat_bound_array,
+            crs=crs,
+            circular=circular,
+            center=center,
+            resolution=resolution,
+        )
+    else:
+        grid_info = RefinedGridInfo(
+            lon_bound_array,
+            lat_bound_array,
+            resolution=resolution,
+        )
+    return grid_info
 
 
 def _regrid_along_grid_dims(regridder, data, grid_x_dim, grid_y_dim, mdtol):
@@ -89,7 +99,7 @@ def _regrid_along_grid_dims(regridder, data, grid_x_dim, grid_y_dim, mdtol):
     return result
 
 
-def _create_cube(data, src_cube, grid_dim_x, grid_dim_y, grid_x, grid_y):
+def _create_cube(data, src_cube, src_dims, tgt_coords, num_tgt_dims):
     """
     Return a new cube for the result of regridding.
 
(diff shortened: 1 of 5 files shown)


esmf_regrid/experimental/io.py
```diff
@@ -52,11 +52,18 @@ def save_regridder(rg, filename):
     regridder_type = rg.__class__.__name__
     if regridder_type == "GridToMeshESMFRegridder":
         src_grid = (rg.grid_y, rg.grid_x)
```
wjbenfold, line 54: I'd suggest that instead of building this and then indexing it, we could just refer to `rg.grid_y` and `rg.grid_x` as we go through this code - I think it will be easier to follow what's going on

esmf_regrid/experimental/io.py
```diff
@@ -52,11 +52,18 @@ def save_regridder(rg, filename):
...
-        src_shape = [len(coord.points) for coord in src_grid]
+        if len(src_grid[0].shape) == 1:
```
wjbenfold, line 55: `src_grid[0].ndim` is equivalent to `len(src_grid[0].shape)` and easier to read I think. This can be changed repeatedly through this function

esmf_regrid/experimental/io.py
```diff
@@ -52,11 +52,18 @@ def save_regridder(rg, filename):
...
+            src_shape = [len(coord.points) for coord in src_grid]
```
wjbenfold, line 56: Using `len(coord.points)` is slightly misleading as it gives the length of the first dimension, rather than the number of items in the array. Given that we've checked that `rg.grid_y` is 1D it doesn't matter here, but `coord.points.size` is equivalent and feels clearer.

esmf_regrid/experimental/io.py
wjbenfold, line 55: Can we assume that one of these being 1D means that the other is? I guess so, but don't know this code well enough to be sure that this contract will always hold.

esmf_regrid/experimental/io.py
```diff
@@ -52,11 +52,18 @@ def save_regridder(rg, filename):
     regridder_type = rg.__class__.__name__
     if regridder_type == "GridToMeshESMFRegridder":
```
wjbenfold, line 53: Given that as far as I can tell we do equivalent steps whichever way around the source and target are, we could build the mesh cube and the grid cube and then assign their names based on what type of regridder we're in? It would reduce code duplication, but might tell it slightly trickier to tell what's going on, so I don't mind strongly either way.

wjbenfold (changes requested):

esmf_regrid/experimental/io.py
stephenworsley: My thinking here is that this might help to keep things clear when we add saving for grid-to-grid regridders. In this case there will be both source and target grids and a potential to reuse this code. Woth that said, I suppose there is room to refactor this code already to reduce code reuse so I might try some refactoring.

esmf_regrid/experimental/io.py
stephenworsley: This should be enforced here: https://github.com/SciTools-incubator/iris-esmf-regrid/blob/266e9ed38588106dec1acc664f2c2fa8bbf476e1/esmf_regrid/schemes.py#L35

esmf_regrid/tests/unit/experimental/io/test_round_tripping.py
```diff
@@ -150,6 +157,25 @@ def test_GridToMeshESMFRegridder_bilinear_round_trip(tmp_path):
...
+    save_regridder(original_rg, filename)
+    loaded_rg = load_regridder(str(filename))
+
+    assert original_rg.grid_x == loaded_rg.grid_x
+    assert original_rg.grid_y == loaded_rg.grid_y
+
+    # Demonstrate regridding still gives the same results.
+    src_data = np.arange(np.product(src.data.shape)).reshape(src.data.shape)
+    src_mask = np.zeros(src.data.shape)
+    src_mask[0, 0] = 1
```
wjbenfold, line 173: I think the following has the same effect:
```
src_data = ma.arange(np.product(src.data.shape)).reshape(src.data.shape)
src_data[0, 0] = ma.masked
```

esmf_regrid/tests/unit/experimental/io/test_round_tripping.py
```diff
@@ -150,6 +157,25 @@ def test_GridToMeshESMFRegridder_bilinear_round_trip(tmp_path):
...
+    src.data = ma.array(src_data, mask=src_mask)
+    # TODO: make this a cube comparison when mesh comparison becomes available.
+    assert np.array_equal(original_rg(src).data, loaded_rg(src).data)
```
wjbenfold, line 176: I think here you're just comparing data values, not whether masks matched. For that I think you need to check the equality of the masks and the data

esmf_regrid/tests/unit/experimental/io/test_round_tripping.py
```diff
@@ -181,17 +203,7 @@ def test_MeshToGridESMFRegridder_round_trip(tmp_path):
...
-    loaded_result = loaded_rg(src)
-    original_result.var_name = loaded_result.var_name
-    original_result.coord("latitude").var_name = loaded_result.coord(
-        "latitude"
-    ).var_name
-    original_result.coord("longitude").var_name = loaded_result.coord(
-        "longitude"
-    ).var_name
-    assert original_result == loaded_result
+    assert np.array_equal(original_rg(src).data, loaded_rg(src).data)
```
wjbenfold, line 206: I think the above masking comments apply here too (and below)

esmf_regrid/tests/unit/experimental/unstructured_scheme/test_MeshToGridESMFRegridder.py
```diff
@@ -346,3 +347,53 @@ def test_resolution():
     lat_band_rg = MeshToGridESMFRegridder(mesh_cube, lat_bands, resolution=resolution)
     assert lat_band_rg.resolution == resolution
     assert lat_band_rg.regridder.tgt.resolution == resolution
+
+
+def test_curvilinear():
+    """
+    Test for :func:`esmf_regrid.experimental.unstructured_scheme.MeshToGridESMFRegridder`.
```
wjbenfold, line 354: This feels like it builds a lot of similar stuff to the test for the other direction - can we share the code that builds the cubes in helper functions?

esmf_regrid/schemes.py
```diff
@@ -49,14 +50,23 @@ def _cube_to_GridInfo(cube):
...
+            lon.points,
+            lat.points,
+            lon_bound_array,
+            lat_bound_array,
+            crs=crs,
+            circular=circular,
+            center=center,
+        )
+    else:
+        grid_info = RefinedGridInfo(
```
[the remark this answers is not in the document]
wjbenfold: This has happened, but I'm not sure I have the power to close a conversation...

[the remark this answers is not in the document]
wjbenfold: This has happened, but I'm not sure I have the power to close a conversation...

esmf_regrid/tests/unit/experimental/io/test_round_tripping.py
```diff
@@ -248,11 +264,10 @@ def test_MeshToGridESMFRegridder_bilinear_round_trip(tmp_path):
...
 
     # Demonstrate regridding still gives the same results.
-    src_data = np.arange(np.product(src.data.shape)).reshape(src.data.shape)
-    src_mask = np.zeros(src.data.shape)
-    src_mask[0] = 1
-    src.data = ma.array(src_data, mask=src_mask)
+    src_data = ma.arange(np.product(src.data.shape)).reshape(src.data.shape)
+    src_data[0] = ma.masked
     assert np.array_equal(original_rg(src).data, loaded_rg(src).data)
+    assert np.array_equal(original_rg(src).data.mask, loaded_rg(src).data.mask)
```
wjbenfold, line 270: It looks like you're doing this full calculation again, rather than just doing it then checking the data and mask.

It also looks like you're not using `src_data` having made it, you're just sending `src` into the regridder without updating its data first?

wjbenfold (changes requested):
I like how this has gone, but I still think we can make it easier to read/maintain:

```
    if regridder_type == "GridToMeshESMFRegridder":
        grid_name = SOURCE_NAME
        mesh_name = TARGET_NAME
    elif regridder_type == "MeshToGridESMFRegridder":
        mesh_name = SOURCE_NAME
        grid_name = TARGET_NAME
    else:
        msg = (
            f"Expected a regridder of type `GridToMeshESMFRegridder` or "
            f"`MeshToGridESMFRegridder`, got type {regridder_type}."
        )
        raise TypeError(msg)

    grid_cube = _standard_grid_cube((rg.grid_y, rg.grid_x), grid_name)

    mesh_coords = rg.mesh.to_MeshCoords(rg.location)
    mesh_data = np.zeros(tgt_mesh_coords[0].points.shape[0])
    mesh_cube = Cube(tgt_data, var_name=mesh_name, long_name=mesh_name)
    for coord in tgt_mesh_coords:
        mesh_cube.add_aux_coord(coord, 0)
```

We then wouldn't need the if/else here too https://github.com/SciTools-incubator/iris-esmf-regrid/blob/d2e22bc97283512a0c4fab131818f4b0c1643b8d/esmf_regrid/experimental/io.py#L147-L150

stephenworsley:
> We then wouldn't need the if/else here too

Actually, I think we can get rid of this anyway because it seems like the bug in iris it's trying to work around has been fixed.

I'm hesitant to organise the cubes by being `mesh_cube` and `grid_cube` because I plan on extending the saving to the grid-to-grid regridder, where there will be two grid cubes. I think the only consistent way to distinguish them is by being source and target.

PR #178, opened 2022-05-04
Allow regridding for projections in non-degree type units

stephenworsley:
Addresses #222 

Some projections, like OSGB, have units which cannot be converted to degrees (i.e. meters) and cause regridding to fail. This PR assumes that where a coordinate system exists, units ought to already be compatible with it.

### DIFF
--- a/esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py
+++ b/esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py
@@ -67,6 +67,4 @@ def test_invalid_tgt_location():
 
 def test_non_degree_crs():
     """Test for coordinates with non-degree units."""
-    expected_sum = 50.86147272655136
-    expected_unmasked = 21
-    _test_non_degree_crs(ESMFAreaWeighted, expected_sum, expected_unmasked)
+    _test_non_degree_crs(ESMFAreaWeighted)
--- a/esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py
+++ b/esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py
@@ -66,7 +66,7 @@ def test_invalid_tgt_location():
 
 
 def test_non_degree_crs():
-    """Test for coordinates with non-degree units"""
+    """Test for coordinates with non-degree units."""
     expected_sum = 50.86147272655136
     expected_unmasked = 21
     _test_non_degree_crs(ESMFAreaWeighted, expected_sum, expected_unmasked)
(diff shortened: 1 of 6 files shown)


trexfeathers:
Using [`pyproj.Proj`](https://pyproj4.github.io/pyproj/stable/api/proj.html) could be an easy way of converting these cases to degrees.

stephenworsley:
> Using [`pyproj.Proj`](https://pyproj4.github.io/pyproj/stable/api/proj.html) could be an easy way of converting these cases to degrees.

This conversion ought to happen at this point:
https://github.com/SciTools-incubator/iris-esmf-regrid/blob/eca3a2a0dc4d5a2e8a2c8ae8535127f719d5fa4c/esmf_regrid/_esmf_sdo.py#L224-L225

I would imagine that cartopy would be calling Proj here anyway. I think that in the case where units are supplied in meters, at this point cartopy describes the projection as being described in meters and is expecting data to be in meters. There is still a case that there could be more to be done to ensure that units are compatible with the coordinate system, but I'm not sure what it would mean for a coordinate to have a unit which disagreed with the coordinate system. 

The approach I've gone for is to treat the coordinate system as the single source of truth for a coordinate where it exists. This seems the simplest way to do things for the time being.

lbdreyer:
Do you plan to add some testing to this PR?

stephenworsley:
> Do you plan to add some testing to this PR?

Yes, I have been thinking about adding tests. I think I would ideally want some OSGB data (or a subset of such data) to work with, since thats where this problem showed up. The closest I can find in iris-test-data seems to be transverse mercator (which i think OSGB is an instance of). If you have any other ideas of data to test this on let me know.

esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py
```diff
@@ -62,3 +63,10 @@ def test_invalid_tgt_location():
     match = "For area weighted regridding, target location must be 'face'."
     with pytest.raises(ValueError, match=match):
         _ = ESMFAreaWeighted(tgt_location="node")
+
+
+def test_non_degree_crs():
+    """Test for coordinates with non-degree units."""
+    expected_sum = 50.86147272655136
+    expected_unmasked = 21
+    _test_non_degree_crs(ESMFAreaWeighted, expected_sum, expected_unmasked)
```
trexfeathers, line 72: This is certainly a difficult one to get right, but to make the code more understandable I think both the setup and expected results need to be in the same place. So either:

- Contain the expected results within `_test_non_degree_crs()`, obviously as some sort of tuple given the various schemes.
- Generate the source and target in the function here, and pass those down to `_test_non_degree_crs()` as more arguments.

trexfeathers (changes requested):
Thanks @stephenworsley, a few comments for you

trexfeathers (approved):

trexfeathers:
Thanks @stephenworsley!

PR #214, opened 2022-10-10
Update to v0.5

stephenworsley:


esmf_regrid/__init__.py
```diff
@@ -1,4 +1,4 @@
 from .schemes import *
 
 
-__version__ = "0.5.dev0"
+__version__ = "0.5"
```
lbdreyer, line 4: Perhaps this should be 
```suggestion
__version__ = "0.5.0"
```
to be consistent with the last release?

lbdreyer (approved):
Looks good!

PR #220, opened 2022-11-21
Benchmarking overhaul and pin Flake8 <6

### LINKED ISSUE #210: Benchmarks failing to run.
## 📰 Custom Issue

See the failing CI.

I expect I baked in some inflexibility as this setup was my first attempt at GitHub-integrated benchmarking. I'd like to replicate the further developed architecture we have in Iris ([see here](https://github.com/SciTools/iris/tree/main/benchmarks)), which would solve the issue and otherwise improve matters. I'll see if I can find time in the coming weeks.

trexfeathers:
A detailed restructure of the benchmarking architecture in line with what has already been done on [SciTools/iris](https://github.com/SciTools/iris). The original architecture included several experiments which we have since iterated and improved within Iris.

Closes #210.


benchmarks/asv_delegated_conda.py
```diff
@@ -0,0 +1,198 @@
+# Copyright Iris contributors
+#
+# This file is part of Iris and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
```
stephenworsley, line 5: While it's probably worth noting that this was copied from Iris, the way this is written implies that this is a file that exists within Iris rather than being copied from there.

stephenworsley (changes requested):
Good work! Just one comment about the license header in one file.

trexfeathers:
See also https://github.com/PyCQA/flake8/pull/1758

trexfeathers:
> Flake8 failure seems possibly related to [PyCQA/flake8#1739](https://github.com/PyCQA/flake8/pull/1739), but I have not yet worked out what in our config is causing the problem.

Happens even without any config at all in `.flake8`, so I reckon somethings wrong. I'll pin it

trexfeathers:
> > Flake8 failure seems possibly related to [PyCQA/flake8#1739](https://github.com/PyCQA/flake8/pull/1739), but I have not yet worked out what in our config is causing the problem.
> 
> Happens even without any config at all in `.flake8`, so I reckon somethings wrong. I'll pin it
> 
> [22a6f63](https://github.com/SciTools-incubator/iris-esmf-regrid/commit/22a6f63fe1ea4990953c0ceb174bca1fcaae96e0)

Should be fixed once PyCQA/flake8-import-order#190 is in a release.

stephenworsley (approved):
Looks good, I guess we can ignore the greetings failure for the time being as that seems to be a seperable issue.

PR #231, opened 2023-01-03
Fix esmpy importing

stephenworsley:


esmf_regrid/_esmf_sdo.py
```diff
@@ -90,14 +98,14 @@ class GridInfo(SDO):
     This class holds information about lat-lon type grids. That is, grids
     defined by lists of latitude and longitude values for points/bounds
     (with respect to some coordinate reference system i.e. rotated pole).
-    It contains methods for translating this information into :mod:`ESMF` objects.
+    It contains methods for translating this information into :mod:`esmpy` objects.
```
trexfeathers, line 101: I don't think your changes to Sphinx domains will work, since those are based on this line:

https://github.com/SciTools-incubator/iris-esmf-regrid/blob/f0781aaff0c657ea878173f6fdcde64dfd113be6/docs/src/conf.py#L118

esmf_regrid/_esmf_sdo.py
```diff
@@ -3,7 +3,15 @@
...
-import ESMF
+
+try:
+    import esmpy
+except ImportError as exc:
+    # Prior to v8.4.0, `esmpy`` could be imported as `ESMF`.
+    try:
+        import ESMF as esmpy  # noqa: N811
+    except ImportError:
+        raise exc
```
trexfeathers, line 14: With such an involved series of steps, might it be worth performing this complex import once in the top level `__init__`, then having all other modules just import from there?

trexfeathers (changes requested):

trexfeathers (changes requested):
I checked out your branch to search for `esmf` throughout the code. There are still Sphinx domains beginning with `ESMF...` in:
- `unstructured_regrid.py`
- `unstructured_scheme.py`
- `schemes.py`

None of these will work now that Sphinx has been configured to use the name `esmpy` in 375adb71b852e04edd2b639730f8f243defdf524.

trexfeathers (changes requested):
There are still plenty in:
- `unstructured_regrid`
- `unstructured_scheme`

trexfeathers (approved):

PR #232, opened 2023-01-04
Drop greetings GHA

bjlittle:
The `greetings` GHA is causing issues, and to be honest it probably doesn't have much value.

I propose that we remove this GHA from the `iris-esmf-regrid` workflow, particularly if it's unstable and causing PR CI to fail.


trexfeathers:
Really had to resist a blind merge for a moment there 😈

Will leave Stephen to have the pleasure

stephenworsley (approved):
Sounds good to me.

PR #236, opened 2023-02-08
Remove flake8 and black Nox sessions.

trexfeathers:
In line with Iris, this PR proposes relying purely on [pre-commit.ci](https://pre-commit.ci/) (already enabled on this repo) for linting tasks, doing away with independently managed Nox sessions.

This reduces duplication, and removes the risk of divergence between the Nox sessions and the pre-commit setup (which recently caused [this CI failure](https://github.com/SciTools-incubator/iris-esmf-regrid/pull/235/checks?check_run_id=11157987258) - different linter versions were insisting on different standards).

Iris' decision-making history:

* Use pre-commit instead of an independent setup:
  SciTools/iris#4181
* Rely purely on pre-commit.ci without an additional Nox session:
  SciTools/iris#4503

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -16,7 +16,6 @@ container:
 
 env:
   # Skip specific tasks by name. Set to a non-empty string to skip.
-  SKIP_LINT_TASK: ""
   SKIP_TEST_TASK: ""
   SKIP_BENCHMARK_TASK: ""
   # Maximum cache period (in weeks) before forcing a new cache upload.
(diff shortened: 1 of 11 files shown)


stephenworsley (changes requested):
I believe black and flake should now not be needed in the requirements files. They are absent from the Iris requirements (removed with this PR https://github.com/SciTools/iris/pull/4181).

Otherwise this looks good.

trexfeathers:
> I believe black and flake should now not be needed in the requirements files. They are absent from the Iris requirements (removed with this PR [SciTools/iris#4181](https://github.com/SciTools/iris/pull/4181)).

Good catch, thanks

trexfeathers:
Looks like the pre-commit config is missing the appropriate subcommands to mimic what was happening in the Nox session. Will look later.

Issue #238, opened 2023-02-20
`dtype` of regridded cube changes after realizing the data


schlunma:
## 🐛 Bug Report

After regridding a cube with `dtype=float32` using `regrid_unstructured_to_rectilinear` and realizing the data the output cube's `dtype` is `float64`. If the data is saved and re-loaded before realizing the data, the `dtype` does **not** change.

## How To Reproduce

```py
import numpy as np

import iris
from iris.coords import DimCoord
from iris.cube import Cube
from iris.experimental.ugrid import PARSE_UGRID_ON_LOAD

import esmf_regrid
from esmf_regrid.experimental.unstructured_scheme import regrid_unstructured_to_rectilinear

print("iris version:", iris.__version__)
print("iris-esmf-regrid version:", esmf_regrid.__version__)

path = iris.sample_data_path('mesh_C4_synthetic_float.nc')
with PARSE_UGRID_ON_LOAD.context():
    cube = iris.load_cube(path)
    print("dtype input cube:", cube.dtype)

lat = DimCoord([10, 20, 30], standard_name='latitude', units='degrees_north')
lon = DimCoord([10, 20, 30], standard_name='longitude', units='degrees_east')
lat.guess_bounds()
lon.guess_bounds()
target = Cube(np.ones((3, 3)), dim_coords_and_dims=[(lat, 0), (lon, 1)])

regridded_cube = regrid_unstructured_to_rectilinear(cube, target)
print("dtype regridded cube before cube.data:", regridded_cube.dtype)

# If the following lines are added, the dtype is still float32 after realizing
# the data
# from pathlib import Path
# x = Path.home() / 'test.nc'
# iris.save(regridded_cube, x)
# regridded_cube = iris.load_cube(x)

regridded_cube.data
print("dtype regridded cube after cube.data:", regridded_cube.dtype)
```

prints

```bash
iris version: 3.4.0
iris-esmf-regrid version: 0.5.0
dtype input cube: float32
dtype regridded cube before cube.data: float32
dtype regridded cube after cube.data: float64
```

## Expected behaviour

The `dtype` should be `float32` for the regridding cube even after realizing the data.

## Environment 
 - OS and version: Red Hat Enterprise Linux 8.5 (Ootpa)
 - `esmf_regrid` version: 0.5.0

bjlittle:
Thanks @schlunma for raising this.

@stephenworsley is on the case and will dip into this asap, cheers :+1:

[...]

PR #239, opened 2023-02-22
Ensure dtype is preserved after regridding

stephenworsley:
Adresses #238.


stephenworsley:
Note: this will still change the dtype if you start with integer data (by design). This does mean that the dask array also will have to be changed to better align with the underlying data in this case. This should be a seperable issue which can be addressed in its own PR.

esmf_regrid/esmf_regridder.py
```diff
@@ -176,12 +176,17 @@ def regrid(self, src_array, norm_type="fracarea", mdtol=1):
         extra_size = max(1, np.prod(extra_shape))
         src_inverted_mask = self.src._array_to_matrix(~ma.getmaskarray(src_array))
         weight_sums = self.weight_matrix @ src_inverted_mask
+        src_dtype = src_array.dtype
+        if np.issubdtype(src_dtype, np.floating):
+            out_dtype = src_dtype
+        else:
+            out_dtype = np.float64
```
pp-mo, line 183: Would it be safer here to use the weights.dtype, if that is float, and only if _that_ is integral default to a fixed floating precision ?
Otherwise, **_any_** integer data will be done in float64, when the weights array might indicate that a lower precision was "good enough" (and presumably cheaper).

esmf_regrid/esmf_regridder.py
```diff
@@ -191,7 +196,7 @@ def regrid(self, src_array, norm_type="fracarea", mdtol=1):
         normalisations = ma.array(normalisations, mask=np.logical_not(tgt_mask))
 
         flat_src = self.src._array_to_matrix(ma.filled(src_array, 0.0))
-        flat_tgt = self.weight_matrix @ flat_src
+        flat_tgt = self.weight_matrix.astype(out_dtype) @ flat_src
```
pp-mo, line 199: Although I'm pretty happy with the other code changes,
I'm not so sure whether **_this_** change is wise in addition ? ...

I note that `flat_src` is derived directly from `source_array`, and _will_ always have the same dtype.

I **_assume_** that the '@' type matrix multiply would behave the same as other "ordinary" arithmetic operations, i.e. the result type is always the higher precision of the two.
However, whether that will always 'uprate' either arg to the precision of the other before the calculation, as this is potentially doing. is rather less certain, I would guess -- we _might_ have specific ways to multiply float32 * float64 without pre-converting the lower-precision array. 

So, I would prefer to _trust_ numpy to handle the operation between these different types appropriately, and not to second-guess it.

Meanwhile, there is also a possibility here that weights is a **_higher_** precision than source_array.  In that case, this line would throw that away.
I think that would also be a mistake :  If I supply a higher-precision weights array, then I'd **_expect_** the operation to perform in higher precision in that case, and to return that higher type.

I think that also alters the scope of the change description here : Don't say that it 'preserves dtype', because I don't think that is always the right answer.  The 'correct' result dtype should be the higher of the weights + and the source-array (and not spuriously increased to float64, as was happening).  In any case, an integral input type is not "preserved".

pp-mo (changes requested):
I'm not convinced that there aren't a few more cases to address here.  I think I would create test examples where both weights + source_array can be integer / float32 / float64, and check that output dtypes are as expected in all those cases.
But I may have missed that dtypes for the weights array are always known, or more limited ?

esmf_regrid/esmf_regridder.py
stephenworsley: I believe the weights tend to be float64 by default, but I suppose this could change if you're using user supplied weights so this is probably worth implementing.

esmf_regrid/esmf_regridder.py
pp-mo: OK, I'm convinced that this now works as expected !

esmf_regrid/esmf_regridder.py
pp-mo: Ok, done!

pp-mo (changes requested):
Basically OK except for a possible tidy in the tests.

esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py
```diff
@@ -74,3 +75,21 @@ def test_invalid_tgt_location():
 def test_non_degree_crs():
     """Test for coordinates with non-degree units."""
     _test_non_degree_crs(ESMFAreaWeighted)
+
+
+@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
pp-mo: @stephenworsley if you think this is too fiddly, I will consider leaving it as-is

esmf_regrid/schemes.py
```diff
@@ -299,6 +301,8 @@ def _map_complete_blocks(src, func, active_dims, out_sizes, *args, **kwargs):
         Dimensions that cannot be chunked.
     out_sizes : tuple of int
         Output size of dimensions that cannot be chunked.
+    dtype : type, optional
+        Type of the output array, if not given, the dtype of src is used.
```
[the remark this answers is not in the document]
pp-mo: Actually no. after offline discussion I think this is a bad suggestion.
Since effectively \*args and \**kwargs contain additional arguments to the wrapped array function, and so should probably *not* contain extra controls for `map_blocks`, since that is a confusingly mixed purpose !

So, I now think this is **best left as it is**.

esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py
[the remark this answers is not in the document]
stephenworsley: see https://github.com/SciTools-incubator/iris-esmf-regrid/pull/239/commits/e341ecf8ee254632d4212a4375d406e72812d5fa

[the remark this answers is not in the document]
pp-mo: OK cool, that does work !
Plus I checked that the pytest output is intelligible in that the way the params print out usefully identifies the individual testcases.
E.G...
```
$ pytest esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py::test_dtype_handling
 ...
esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py::test_dtype_handling[src_tgt_types0-float32] PASSED                                                                                [ 12%]
esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py::test_dtype_handling[src_tgt_types0-float64] PASSED                                                                                [ 25%]
esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py::test_dtype_handling[src_tgt_types1-float32] PASSED                                                                                [ 37%]
esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py::test_dtype_handling[src_tgt_types1-float64] PASSED                                                                                [ 50%]
esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py::test_dtype_handling[src_tgt_types2-float32] PASSED                                                                                [ 62%]
esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py::test_dtype_handling[src_tgt_types2-float64] PASSED                                                                                [ 75%]
esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py::test_dtype_handling[src_tgt_types3-float32] PASSED                                                                                [ 87%]
esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py::test_dtype_handling[src_tgt_types3-float64] PASSED                                                                                [100%]
 ...
itpp@vld321.cmpd1.metoffice.gov.uk $ pytest esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py::test_cube_regrid

esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py::test_cube_regrid[grid-grid] PASSED                                                                                                [ 20%]
esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py::test_cube_regrid[grid-mesh] PASSED                                                                                                [ 40%]
esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py::test_cube_regrid[grid-just_mesh] PASSED                                                                                           [ 60%]
esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py::test_cube_regrid[mesh-grid] PASSED                                                                                                [ 80%]
esmf_regrid/tests/unit/schemes/test_ESMFAreaWeighted.py::test_cube_regrid[mesh-mesh] PASSED                                                                                                [100%]
```

I have found from past experience with pytest parameters, that is really worth paying attention to this, to make sure you can work out which testcase is which !
If necessary, you can use string params and decode them, or "ids=" as strings describing the parameter values.

pp-mo (approved):

Issue #227, opened 2022-12-16
ESMPy API change in v8.4.0


bouweandela:
Note that starting with [ESMF 8.4.0, ESMPy should be imported with `import esmpy` instead of `import ESMF`](https://earthsystemmodeling.org/static/releases.html). This would make the current version incompatible with the latest version of ESMF unfortunately.

zklaus:
As an aside, @bouweandela wrote a nifty little solution for ESMValTool that you might copy in ESMValGroup/ESMValCore#1876.

pletzer:
This is also mentioned in #229

valeriupredoi:
hey guys, this would be very nice to have since this also blocks support for Python 3.11 which would be nice to have also - if you want me to I can open a PR with it? :beer:

valeriupredoi:
I am being goody two-shoes so I opened a PR :grin: https://github.com/SciTools-incubator/iris-esmf-regrid/pull/240

valeriupredoi:
This was a good iisue, RIP good issue 😢 Also, this kind of auto-closer bot, like iris has one too, may be harmful, if a human is not checking what it closed

PR #240, opened 2023-02-23
Fix non-functional `import ESMF` after `esmpy>=8.4`

valeriupredoi:
This is @bouweandela 's elegant solution from https://github.com/ESMValGroup/ESMValCore/pull/1876 implemented here so `iris-esmf-regrid` can support the newer esmpy, and implicitly allow support for Python=3.11, implicitly allowing us to have Python=3.11 too :grin: 

Closes https://github.com/SciTools-incubator/iris-esmf-regrid/issues/227

### DIFF
--- a/esmf_regrid/_esmf_sdo.py
+++ b/esmf_regrid/_esmf_sdo.py
@@ -3,7 +3,14 @@
 from abc import ABC, abstractmethod
 
 import cartopy.crs as ccrs
-import ESMF
+try:
+    import esmpy
+except ImportError as exc:
+    # Prior to v8.4.0, `esmpy`` could be imported as `ESMF`.
+    try:
+        import ESMF as esmpy  # noqa: N811
+    except ImportError:
+        raise exc
 import numpy as np
 import scipy.sparse
 
--- a/esmf_regrid/_esmf_sdo.py
+++ b/esmf_regrid/_esmf_sdo.py
@@ -3,6 +3,7 @@
 from abc import ABC, abstractmethod
 
 import cartopy.crs as ccrs
+
 try:
     import esmpy
 except ImportError as exc:
--- a/esmf_regrid/_esmf_sdo.py
+++ b/esmf_regrid/_esmf_sdo.py
@@ -3,7 +3,6 @@
 from abc import ABC, abstractmethod
 
 import cartopy.crs as ccrs
-
 try:
     import esmpy
 except ImportError as exc:
--- a/esmf_regrid/_esmf_sdo.py
+++ b/esmf_regrid/_esmf_sdo.py
@@ -1,4 +1,4 @@
-"""Provides representations of ESMF's Spatial Discretisation Objects."""
+"""Provides representations of ESMF/esmpy's Spatial Discretisation Objects."""
 
 from abc import ABC, abstractmethod
 
@@ -35,7 +35,7 @@ class SDO(ABC):
     def make_esmf_field(self):
         """Return an ESMF field representing the spatial discretisation object."""
         sdo = self._make_esmf_sdo()
-        field = ESMF.Field(sdo, **self._field_kwargs)
+        field = esmpy.Field(sdo, **self._field_kwargs)
         return field
 
     @property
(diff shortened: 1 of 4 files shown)


valeriupredoi:
hey guys - this **should** be ready, but there are two things to look at:

- there are still a whole bunch of instances of ESMF in docstrings - I have not changed those to esmpy since I think they should still reference ESMF, since that's the original package
- I am a bit confused as to why the pre-commit test is failing, I ran `flake8` in my repo and all is fine :confused: 

Cheers :beer:

valeriupredoi:
> Thanks for the PR, I'll look into the pre-commit business.

awesome, cheers muchly @stephenworsley :beer:

stephenworsley:
Looks like this was a case of black and flake8 disagreeing. I've turned off black formatting for these lines of code and that seems to have fixed it.

valeriupredoi:
great! Sorry about that, I don't use `black` :black_flag:

dhohn:
duplicate of the following? 
 https://github.com/SciTools-incubator/iris-esmf-regrid/pull/231

That PR had a different target tho...

valeriupredoi:
that's indeed what we're after here too - with the added bonus of having docstrings changed with ESMF -> esmpy too. But that's a different branch, as you mention - not sure if the devs would rather open a PR off that branch to `main` or accept this PR - whatever's best for them methinks :beer:

stephenworsley:
Okay, so it turns out the situation is actually a bit more complex... So not only has the name changed from ESMF to esmpy, the installation requirements have also changed. You now need to set an environment variable `ESMFMKFILE` to import esmpy (https://earthsystemmodeling.org/esmpy_doc/release/latest/ESMPy.pdf#section.3.4). You can see the errors you get with this approach in the tests for this PR #241. I'm not sure how to tell what the appropriate `ESMFMKFILE` would be in the CI or in general so I'm not sure how to solve this one right now...

dhohn:
> Okay, so it turns out the situation is actually a bit more complex... So not only has the name changed from ESMF to esmpy, the installation requirements have also changed. You now need to set an environment variable `ESMFMKFILE` to import esmpy (https://earthsystemmodeling.org/esmpy_doc/release/latest/ESMPy.pdf#section.3.4). You can see the errors you get with this approach in the tests for this PR #241. I'm not sure how to tell what the appropriate `ESMFMKFILE` would be in the CI or in general so I'm not sure how to solve this one right now...

Are you installing into an active conda env? If so I had the same issue: https://github.com/conda-forge/esmf-feedstock/issues/91#issuecomment-1350821853

A `deactivate` + `activate` of the conda env should do the trick in that case.

stephenworsley:
> Are you installing into an active conda env? If so I had the same issue: [conda-forge/esmf-feedstock#91 (comment)](https://github.com/conda-forge/esmf-feedstock/issues/91#issuecomment-1350821853)
> 
> A `deactivate` + `activate` of the conda env should do the trick in that case.

Interesting, I'll have to figure out how to do that in the CI

zklaus:
Is it a problem in the CI? It shouldn't be: The environment will be setup first and then activated. The problem only arises if one is inside an activated environment and then installs a new version of the esmpy package.

valeriupredoi:
@zklaus is right - also, looking at your [CI test config](https://github.com/SciTools-incubator/iris-esmf-regrid/blob/main/.cirrus.yml) it looks to me like all the stuff is happening in the `(base)` conda environment, which is not a good idea, one needs to create separate envs on top of the base env. I am not too familiar with Cirrus CI, so I might be wrong though :beer:

stephenworsley:
With #228 merged in I believe this should be sorted.

valeriupredoi:
hi @stephenworsley cheers! #228 folds in all the needed changes brought in by this PR too - slightly misleading since that PR is called CI/lockfiles. Should I close this then? :beer:

valeriupredoi:
sure thing! Cheers @stephenworsley - do you guys need any help from me with the new release that'd include the new changes?

valeriupredoi:
awesome stuff! Will do, heading there now :beer: Let me know if any help's needed with the conda packaging etc - will be glad to help!

PR #242, opened 2023-03-06
Fix curvilinear lazy regridding

stephenworsley:
Curvilinear regridding is currently bugged when the data is lazy. I think this might also be fixed in #198, but this should also fix it in the short term. #198 will have to be refactored in light of this.

### DIFF
--- a/esmf_regrid/experimental/unstructured_scheme.py
+++ b/esmf_regrid/experimental/unstructured_scheme.py
@@ -220,6 +220,7 @@ def _regrid_unstructured_to_rectilinear__perform(src_cube, regrid_info, mdtol):
     if len(grid_x.shape) == 1:
         chunk_shape = (len(grid_x.points), len(grid_y.points))
     else:
+        # Due to structural reasons, the order here must be reversed.
         chunk_shape = grid_x.shape[::-1]
     new_data = _map_complete_blocks(
         src_cube,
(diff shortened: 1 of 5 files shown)


esmf_regrid/tests/unit/schemes/test_regrid_rectilinear_to_rectilinear.py
```diff
@@ -177,6 +177,70 @@ def test_laziness():
...
+
+    src_data = np.empty([h, src_lats, t, src_lons, e])
+    src_data[:] = np.arange(t * h * e).reshape([h, t, e])[
+        :, np.newaxis, :, np.newaxis, :
+    ]
+
+    src_data = np.empty([h, src_lats, t, src_lons, e])
+    src_data[:] = np.arange(t * h * e).reshape([h, t, e])[
+        :, np.newaxis, :, np.newaxis, :
+    ]
```
lbdreyer, line 224: These two lines are duplicated from above

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -220,7 +220,7 @@ def _regrid_unstructured_to_rectilinear__perform(src_cube, regrid_info, mdtol):
     if len(grid_x.shape) == 1:
         chunk_shape = (len(grid_x.points), len(grid_y.points))
     else:
-        chunk_shape = grid_x.shape
+        chunk_shape = grid_x.shape[::-1]
```
lbdreyer, line 223: Is this worth a comment explaining why this is necessary?

lbdreyer, line 223: I can't find where this is tested that it is preserving laziness?

lbdreyer: As discussed offline, this change doesn't "preserve laziness", but rather that it is a fix required as the lazy data was failing.

lbdreyer (approved):

lbdreyer:
Thanks @stephenworsley

PR #252, opened 2023-04-05
check-manifest and pep621

bjlittle:
This PR introduces `MANIFEST.in` checking to ensure that the PyPI source distribution and binary wheels are packaged with the expected payload.

Additionally, also applies [pep-621](https://peps.python.org/pep-0621/) to modernize `setuptools` and migrate from the `setup.cfg` to `pyproject.toml`.

Note that, `esmf` is not available on PyPI, therefore a `pip install esmf_regrid` will require the `esmf` package to be installed via other means e.g., `conda`.

Reference:
- https://packaging.python.org/en/latest/guides/using-manifest-in/
- https://github.com/mgedmin/check-manifest
- https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#


stephenworsley (approved):
Looks good.

PR #258, opened 2023-04-13
Add env vars to sphinx config

### LINKED ISSUE #251: readthedocs build is failing
## 🐛 Bug Report

The `readthedocs` build has been failing since the last 3 months, we should:

- [x] fix this
- [x] add a RDT badge to the `README.md` so that it easily alerts developers

zklaus:
Readthedocs doesn't `activate` Conda environments and also doesn't use `conda run`. As a consequence, crucial environment variables are not set, preventing the successful building of the documentation.

This PR adds the environment variables in Sphinx' `conf.py`, providing a workaround for that problem. This might need to be updated in the future if either different variables are needed (e.g. Proj recently moved from `PROJ_LIB` to `PROJ_DATA`), or if Readthedocs addresses this, e.g. in readthedocs/readthedocs.org#5339.

Fixes #251

### DIFF
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -18,8 +18,8 @@ source_code_root = (Path(__file__).parents[2]).absolute()
 sys.path.append(str(source_code_root))
 
 on_rtd = os.environ.get("READTHEDOCS", None) == "True"
-rtd_version = os.environ.get("READTHEDOCS_VERSION", "latest")
 if on_rtd:
+    rtd_version = os.environ.get("READTHEDOCS_VERSION", "latest")
     rtd_project = os.environ.get("READTHEDOCS_PROJECT")
     rtd_conda_prefix = f"/home/docs/checkouts/readthedocs.org/user_builds/{rtd_project}/conda/{rtd_version}"
     os.environ["ESMFMKFILE"] = f"{rtd_conda_prefix}/lib/esmf.mk"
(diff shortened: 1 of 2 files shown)


bjlittle:
Hey @zklaus, thanks for pushing this along 💯 🍻 

I just approved your request for access to RTD as a maintainer, so you're unblocked now 👍

docs/src/conf.py
```diff
@@ -16,6 +17,15 @@
 source_code_root = (Path(__file__).parents[2]).absolute()
 sys.path.append(str(source_code_root))
 
+on_rtd = os.environ.get("READTHEDOCS", None) == "True"
+rtd_version = os.environ.get("READTHEDOCS_VERSION", "latest")
```
stephenworsley, line 21: It seems a little inconsistent having this line outside the `if on_rtd:` logic while `rtd_project` is set within given that both of these are only used within that scope.

stephenworsley (changes requested):
Just one small nitpick but otherwise this looks good.

docs/src/conf.py
zklaus: True. For us, i.e. in ESMValTool, we do this because there are a few links between projects where we also want to link to the latest rtd_version when the documentation is built locally. But perhaps this gets in the way here?

stephenworsley: I suppose that makes sense if it's following a pattern. With that said, I can't imagine a case where we would add another project in this code so I think it ought to be fine to change it.

zklaus: Done.

stephenworsley (approved):
Looks good.

PR #270, opened 2023-05-19
Introduce enums to replace free text user inputs. 

### LINKED ISSUE #263: Use `enum` around the codebase
There are several examples in iris-esmf-regrid of 'modes', e.g. `conservative`/`bilinear`, `source`/`target`. So there are lots of checks for `== "conservative"` etcetera. These are vulnerable to spelling mistakes, and force object signatures to accept free text when there are only a few valid values.

These are prime cases for using Python's [`enum`](https://docs.python.org/3/library/enum.html) module instead. Once in place, code will be easier to follow and to modify, with fewer mistakes.

ESadek-MO:
Closes #263.

The introduction of enums in place of user inputs makes future edits easier for developers, and lowers chance of user error.

The current intention is to include the enum classes as a Class Attribute, so that users aren't required to import them independently. 

- [x]  esmf-regridder
- [x] schemes
   - [x] different structure and usage 
- [x] unstructured-regrid
- [x] unstructured-scheme?
-----
- [ ] tgt/src
- [ ] tests

### DIFF
--- a/esmf_regrid/__init__.py
+++ b/esmf_regrid/__init__.py
@@ -7,7 +7,8 @@ except ImportError as exc:
     except ImportError:
         raise exc
 
-from .schemes import *
 
+from .schemes import *
+from ._constants import Constants
 
 __version__ = "0.6.0"
--- a/esmf_regrid/__init__.py
+++ b/esmf_regrid/__init__.py
@@ -7,8 +7,9 @@ except ImportError as exc:
     except ImportError:
         raise exc
 
-
+# constants is used within schemes, so needs to imported first
+from ._constants import *
 from .schemes import *
-from ._constants import Constants
+
 
 __version__ = "0.6.0"
--- a/esmf_regrid/__init__.py
+++ b/esmf_regrid/__init__.py
@@ -8,7 +8,7 @@ except ImportError as exc:
         raise exc
 
 # constants is used within schemes, so needs to imported first
-from ._constants import *
+from ._constants import Constants
 from .schemes import *
 
 
--- a/esmf_regrid/esmf_regridder.py
+++ b/esmf_regrid/esmf_regridder.py
@@ -64,6 +64,8 @@ class Regridder:
         NEAREST = esmpy.RegridMethod.NEAREST_STOD
 
     NormType = Enum("NormType", ["FRACAREA", "DSTAREA"])
+    # used in other files, placed here to have them all in one place
+    Location = Enum("Location", ["FACE", "NODE"])
 
     def __init__(self, src, tgt, method=Method.CONSERVATIVE, precomputed_weights=None):
         """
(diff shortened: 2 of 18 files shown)


esmf_regrid/esmf_regridder.py
```diff
@@ -76,8 +84,9 @@ def __init__(self, src, tgt, method="conservative", precomputed_weights=None):
...
             Data output by this regridder will be a :class:`numpy.ndarray` whose
             shape is compatible with ``tgt``.
-        method : str
-            Either "conservative", "bilinear" or "nearest". Corresponds to the :mod:`esmpy` methods
+        method : :class:`Regridder.Method`
+            Either Regridder.Method.CONSERVATIVE, Regridder.Method.BILINEAR or Regridder.Method.NEAREST.
+            Corresponds to the :mod:`esmpy` methods
             :attr:`~esmpy.api.constants.RegridMethod.CONSERVE`,
             :attr:`~esmpy.api.constants.RegridMethod.BILINEAR` or
             :attr:`~esmpy.api.constants.RegridMethod.NEAREST_STOD` used to calculate weights.
```
trexfeathers, line 92: One of the appeals of using `Enum` is being able to avoid listing things within other docstrings. ``:class:`Regridder.Method` `` renders as a nice link to your new enumeration; we could put the bulk of the documentation within `Method` instead, just leaving basic info behind.

image

esmf_regrid/esmf_regridder.py
```diff
@@ -57,7 +58,14 @@ def _weights_dict_to_sparse_array(weights, shape, index_offsets):
...
 
-    def __init__(self, src, tgt, method="conservative", precomputed_weights=None):
+    class Method(Enum):
+        CONSERVATIVE = esmpy.RegridMethod.CONSERVE
+        BILINEAR = esmpy.RegridMethod.BILINEAR
+        NEAREST = esmpy.RegridMethod.NEAREST_STOD
+
+    NormType = Enum("NormType", ["FRACAREA", "DSTAREA"])
+
+    def __init__(self, src, tgt, method=None, precomputed_weights=None):
```
trexfeathers, line 68: I know `self.Method.CONSERVATIVE` doesn't work, but I don't get any linting errors when I try:

```suggestion
    def __init__(self, src, tgt, method=Method.CONSERVATIVE, precomputed_weights=None):
```

esmf_regrid/esmf_regridder.py
trexfeathers:
```suggestion
        method : :class:`Regridder.Method`
            The method to be used to calculate weights.
```

esmf_regrid/__init__.py
```diff
@@ -7,6 +7,8 @@
     except ImportError:
         raise exc
 
+# constants is used within schemes, so needs to imported first
+from ._constants import *
 from .schemes import *
```
bjlittle, line 12: The adoption of the `from <x> import *` pattern is generally discouraged, particularly if the source module (`<x>`) doesn't have a `__all__` defined.

There is plenty blurb out there backing up this argument e.g., see
- https://stackoverflow.com/questions/2360724/what-exactly-does-import-import
- https://stackoverflow.com/questions/44834/what-does-all-mean-in-python

PR #273, opened 2023-05-23
Update to v0.7

stephenworsley:
To be merged only after #272


pp-mo (approved):
Also LGTM !

Assuming, as I think, there is as yet no automation of versioning, and the version is solely controlled by [setting `__version__` in `/__init__.py`](https://github.com/SciTools-incubator/iris-esmf-regrid/blob/main/esmf_regrid/__init__.py#L13) ?
(I did search and it seems consistent).

pp-mo:
Thanks @stephenworsley 
All ok AFAICT 
If you take out of draft, I will merge !

stephenworsley:
@pp-mo Yes, we're looking to add automatic versioning later.

Issue #245, opened 2023-03-28
GridToMeshESMFRegridder: Allow use of mesh objects for target mesh


arjclark:
## ✨ Feature Request
It would be incredibly useful when generating regridders for GridToMeshESMFRegridder to be able to use mesh objects rather than requiring a cube for the target.

## Motivation
This is motivated by the need to generate ancillary files for running LFRic. In NGMS an LFRic mesh generator is used to provide mesh files for how the model should be set up. These are loadable by iris using the `load_mesh` routine and all seem correct, etc. However, when it comes to constructing a regridder you can't use them in operations like: `rg = GridToMeshESMFRegridder(regridme, targetmesh)` as you get the error: `AttributeError: 'Mesh' object has no attribute 'mesh'`. Thus, in order to use the mesh we would then need to write code to generate an empty cube to attach the mesh to. This is redundant and unnecessary boilerplate.

@stephenworsley

stephenworsley:
closed by #289

PR #289, opened 2023-07-10
Add option to initialise from mesh

stephenworsley:
Adresses #245


HGWright (approved):
Looks great, tests all passing, thanks for the changes we talked about!

PR #293, opened 2023-07-30
updated for enums

### LINKED ISSUE #263: Use `enum` around the codebase

ESadek-MO:
Closes #263 

Introducing enums into iris esmf_regrid, to make code changes easier, user input more directed and prevent unexpected inputs.


stephenworsley (changes requested):
Looking good, I think this adds some good structure internally and makes the connection to ESMPy clearer. I think it could be possible to do this in a more lenient way with just a little more effort.

esmf_regrid/constants.py
```diff
@@ -0,0 +1,18 @@
...
+    """Encompassing class for best practice import."""
+
+    class Method(Enum):
+        """holds enums for Method values."""
+
+        CONSERVATIVE = esmpy.RegridMethod.CONSERVE
+        BILINEAR = esmpy.RegridMethod.BILINEAR
+        NEAREST = esmpy.RegridMethod.NEAREST_STOD
+
+    NormType = Enum("NormType", ["FRACAREA", "DSTAREA"])
```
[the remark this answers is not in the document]
ESadek-MO: I'll happily let you veto, but I initially decided against this because it never directly uses it, whereas Method does.

esmf_regrid/experimental/unstructured_scheme.py
```diff
@@ -413,6 +404,7 @@ def regrid_unstructured_to_unstructured(
         A new :class:`~iris.cube.Cube` instance.
 
     """
```
stephenworsley, line 406: The `method` keyword for this function should have the default `Constants.Method.CONSERVATIVE`

esmf_regrid/schemes.py
```diff
@@ -457,37 +458,35 @@ def copy_coords(src_coords, add_method):
...
     else:
         mesh = cube_or_mesh.mesh
         location = cube_or_mesh.location
     if mesh is None:
         raise ValueError(f"The {src_or_tgt} cube is not defined on a mesh.")
-    if method == "conservative":
+    if method == Constants.Method.CONSERVATIVE:
         if location != "face":
             raise ValueError(
                 f"Conservative regridding requires a {src_or_tgt} cube located on "
```
stephenworsley, line 486: For the sake of consistency, it's worth considering changing "Conservative" to "`{method}`". I think it's also worth either changing `{method}` for `{method.name}` for the sake of clarity and readability or else rewriting this error message so that the formatting of `{method}` reads less awkwardly (e.g. "regridding with the method {method} requires...").

stephenworsley (changes requested):
Almost there, just a couple minor things and this is looking good to go.

stephenworsley (approved):
Good work! Thanks for soldiering on with this one @ESadek-MO.

PR #302, opened 2023-08-30
Extending stale GHA time

HGWright:
The stale bot on this repo is currently very aggressive which leads to lots of relevent issues and pr's getting tagged with the stale label. This requires more intervention to remove the label and mark them as still relevent. By extending the time before issues are marked as stale we will reduce the manual intervention required.


stephenworsley (approved):
Looks good.

PR #338, opened 2024-02-21
Fix usage of map_blocks

stephenworsley:
See https://github.com/SciTools/iris/pull/5767

### DIFF
--- a/esmf_regrid/schemes.py
+++ b/esmf_regrid/schemes.py
@@ -327,7 +327,7 @@ def _map_complete_blocks(src, func, active_dims, out_sizes, *args, **kwargs):
     if num_out == 1:
         out_chunks[sorted_dims[0]] = out_sizes[0]
     else:
-        for dim, size in zip(dims, out_sizes):
+        for dim, size in zip(active_dims, out_sizes):
             # Note: when mapping 2D to 2D, this will be the only alteration to
             # out_chunks, the same as iris._lazy_data.map_complete_blocks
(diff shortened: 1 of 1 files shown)


esmf_regrid/schemes.py
```diff
@@ -276,7 +275,7 @@ def _regrid_along_dims(regridder, data, dims, num_out_dims, mdtol):
     return result
 
 
-def _map_complete_blocks(src, func, dims, out_sizes):
+def _map_complete_blocks(src, func, active_dims, out_sizes, *args, **kwargs):
```
bjlittle, line 278: @stephenworsley Is your inclusion of `*args` here speculative of future `func` API?

At the moment `*args` doesn't seem to be used ... just curious.

bjlittle (changes requested):
@stephenworsley Looks good 👍 

Just a simple question to service, also I guess existing testing covers this, right?

I wouldn't expect you to add tests to ensure that `partial` isn't used 😜

esmf_regrid/schemes.py
stephenworsley: yeah, I suppose this is speculative for the time being, though the idea here is that I wanted to match the signature used by `map_blocks`. I decided to preserve both args and kwargs for the sake of completeness.

stephenworsley: This also matches the pattern I used in iris here https://github.com/SciTools/iris/pull/5767 which preserve args for the same reasons.

PR #357, opened 2024-05-01
Extend regridder saving/loading to all regridders

### LINKED ISSUE #318: Support saving/loading of `ESMFAreaweightedRegridder`, `ESMFBilinearRegridder` and `ESMFNearestRegridder`.
Support the saving/loading of all other regridders. Retain the ability to load a regridder previously saved as a `MeshToGridESMFRegridder` or `GridToMeshESMFRegridder`. These ought to be loaded as an `ESMFAreaweightedRegridder`, `ESMFBilinearRegridder` or `ESMFNearestRegridder`.

stephenworsley:
Closes #318


esmf_regrid/experimental/io.py
```diff
@@ -43,6 +57,49 @@ def _add_mask_to_cube(mask, cube, name):
...
+        for coord in cube.mesh.coords():
+            coord.var_name = None
+        for con in cube.mesh.connectivities():
+            con.var_name = None
+    return cube
+
+
 def save_regridder(rg, filename):
     """
     Save a regridder scheme instance.
```
pp-mo, line 105: Outside the changes so I can't comment directly, but the following section is now out of date.
```
 Saves either a
    :class:`~esmf_regrid.experimental.unstructured_scheme.GridToMeshESMFRegridder`
    or a
    :class:`~esmf_regrid.experimental.unstructured_scheme.MeshToGridESMFRegridder`.
    Parameters
    ----------
    rg : :class:`~esmf_regrid.experimental.unstructured_scheme.GridToMeshESMFRegridder` or :class:`~esmf_regrid.experimental.unstructured_scheme.MeshToGridESMFRegridder`
        The regridder instance to save.
```

Needs updating to reference the correct supported regridder types

esmf_regrid/experimental/io.py
```diff
@@ -43,6 +57,49 @@ def _add_mask_to_cube(mask, cube, name):
         cube.add_aux_coord(mask_coord, list(range(cube.ndim)))
 
 
+@contextmanager
+def _managed_var_name(src_cube, tgt_cube):
```
pp-mo, line 61: I don't really see the purpose of defining this as a _context manager_, 
since it anyway only makes some _permanent_ changes to the arguments (mutable)?
So, it has no 'yield', and no code to "undo afterwards".

So, instead of the style ..
```
with change_cubes(cubes):
   operate(cubes)
```
You could simply
```
change_cubes(cubes)
operate(cubes)
```

esmf_regrid/experimental/io.py
```diff
@@ -43,6 +57,49 @@ def _add_mask_to_cube(mask, cube, name):
...
+            coord.var_name = "_".join([TARGET_NAME, "mesh", coord.name()])
+        yield None
+    finally:
+        for coord, var_name in zip(src_mesh_coords, src_coord_names):
+            coord.var_name = var_name
+        for coord, var_name in zip(tgt_mesh_coords, tgt_coord_names):
+            coord.var_name = var_name
+
+
+def _clean_var_names(cube):
```
pp-mo, line 90: Again, because this takes a cube + returns a cube, it gives the impression that it creates + returns a new cube, leaving the original argument alone.  I.E. it appears to have a functional style.
But in fact, it makes _permanent_ changes to the argument cube, and then just returns that.

I think that, if you _want_ to create a new cube (and leave the argument alone), then you should _copy_ it. 
Or if not, you can just define and use this routine as modifying its argument (as also suggested above for `_managed_var_name` ) : no need to return a cube.

esmf_regrid/experimental/io.py
```diff
@@ -158,17 +258,14 @@ def _standard_grid_cube(grid, name):
...
+    # Save cubes while ensuring var_names do not conflict for the sake of consistency.
+    with _managed_var_name(src_cube, tgt_cube):
         cube_list = CubeList([src_cube, tgt_cube, weights_cube, weight_shape_cube])
-    elif regridder_type == "MeshToGridESMFRegridder":
-        cube_list = CubeList([tgt_cube, src_cube, weights_cube, weight_shape_cube])
 
-    for cube in cube_list:
-        cube.attributes = attributes
+        for cube in cube_list:
+            cube.attributes = attributes
```
pp-mo, line 266: I presume that the purpose of assigning the same attributes to all cubes is to ensure that they are all made global (file) attributes?
It would be nice to leverage the new global+local attributes management to make that clearer here,
E.G. I would suggest you could just 
`src_cube.attributes = CubeAttrsDict(globals=attributes)`

However, possibly we can't do this, if we intend to still support Iris versions <3.8 ??

esmf_regrid/tests/unit/experimental/io/test_round_tripping.py
```diff
@@ -110,38 +121,57 @@ def _make_mesh_to_grid_regridder(
...
-        use_src_mask=use_src_mask,
-        use_tgt_mask=use_tgt_mask,
+        **kwargs,
     )
     return rg, src
 
 
+def _compare_ignoring_var_names(x, y):
+    old_var_name = x.var_name
+    x.var_name = y.var_name
```
pp-mo, line 142: Is this actually good enough -- what about the var-names of coordinates, or the mesh (for instance) ?

Obviously though, it does work for now.

pp-mo (changes requested):
I'm not really competent to comment on all the mechanics here.
But I have raised some 'style' points, including docstrings that seem to need changing.

Also, I think it probably needs a changelog entry ?

esmf_regrid/experimental/io.py
stephenworsley: The "undo afterwards" code should be in the `finally:` part. I've added some tests to show that this does actually revert back the changed to the original cube. See https://github.com/SciTools-incubator/iris-esmf-regrid/pull/357/commits/05209672d69270925eacbe4a7772d05288d07adf

esmf_regrid/experimental/io.py
stephenworsley: The problem with copying a cube is that we've set up cube copying in such a way that the mesh on a copy is the same mesh as the original, so if you edit the mesh on the copy of a cube you are editing the mesh on the orignal cube. Also, this function is only called on the loaded cube so editing it in place makes more sense, I can change the return line though.

esmf_regrid/experimental/io.py
pp-mo: OK I clearly hadn't read this carefully enough.  There **_is_** a `yield` (!)

esmf_regrid/experimental/io.py
pp-mo: 👍

esmf_regrid/tests/unit/experimental/io/test_round_tripping.py
stephenworsley: I think the idea is that this is for comparing coordinates, which would only have the surface level var name. comparing meshes is a bit awkward currently and so is a bit of a blind spot across iris_esmf_regrid (as noted in various code comments). perhaps this can be clarified in the naming of this helper function?

esmf_regrid/experimental/io.py
stephenworsley: I believe that is the idea, this should put attributes in a more readable place in the file (and keep the structure more similar to ESMFs file structure). I think that would be a good idea for a future change, but is probably out of scope for this PR since it does change the dependencies.

esmf_regrid/tests/unit/experimental/io/test_round_tripping.py
pp-mo: OK I didn't quite grasp why the coordinate names might not also be changed in the save process,
but if you're happy then it doesn't really matter.

pp-mo (approved):

pp-mo:
OK I think this is all good now

PR #418, opened 2024-10-03
Authorise the lock file pull request using the SciTools App token

trexfeathers:
I'm afraid that by far the easiest way to test this is to merge and then run Workflow Dispatch. If we discover problems then I will put in the work to set up a testing instance for myself, but I'd rather do that in the second instance.

### DIFF
--- a/.github/workflows/refresh-lockfiles.yml
+++ b/.github/workflows/refresh-lockfiles.yml
@@ -25,15 +25,26 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+
       - name: install nox
         run: |
           source $CONDA/bin/activate base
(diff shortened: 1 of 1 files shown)


.github/workflows/refresh-lockfiles.yml
trexfeathers, line 1: Other examples:

- [cf-units](https://github.com/SciTools/cf-units/blob/81b9d5bfcdaa369461b6834f01b46bea4378b5f5/.github/workflows/ci-locks.yml#L114-L126)
- [workflows](https://github.com/SciTools/workflows/blob/399409ddf520c7159a807a6768db785785843d2e/.github/workflows/refresh-lockfiles.yml#L79-L90) (which I guess we'll want to use eventually)

stephenworsley (approved):
Looks good.

PR #426, opened 2024-10-30
Numpy 2 fix

stephenworsley:
Applies fixes on top of #420


pp-mo (approved):
BTW I really wish they _hadn't_ done this.  But not for us to say !

Issue #429, opened 2024-11-05
TypeError: 'ParseUGridOnLoad' object is not callable with iris<3.10 when using load_regridder


rhyspoulton:
## 🐛 Bug Report

I am attempting to use the `load_regridder` method with the `ESMFBilinear` regridder method, butI am encountering the following issue:
```python
Traceback (most recent call last):
  File ".../esmf-test/test_regrid.py", line 71, in <module>
    loaded_regridder = load_regridder("saved_regridder.nc")
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../esmf_regrid/experimental/io.py", line 292, in load_regridder
    with _load_context():
         ^^^^^^^^^^^^^^^
TypeError: 'ParseUGridOnLoad' object is not callable
```

This seems to be because `iris.experimental.ugrid.PARSE_UGRID_ON_LOAD` is being incorrectly called. According to the [iris documentation](https://scitools-iris.readthedocs.io/en/v3.8.1/generated/api/iris.experimental.ugrid.load.html#iris.experimental.ugrid.load.ParseUGridOnLoad) it should be called in the following way:
```python
with PARSE_UGRID_ON_LOAD.context():
    my_cube_list = iris.load([my_file_path, my_file_path2],
                             constraint=my_constraint,
                             callback=my_callback)
```

## How To Reproduce

Execute the following script with iris<3.10
```python
import iris
from esmf_regrid.experimental.io import load_regridder, save_regridder
from esmf_regrid.schemes import ESMFBilinear

# Load test data
global_air_temp = iris.load_cube(iris.sample_data_path('air_temp.pp'))
rotated_psl = iris.load_cube(iris.sample_data_path('rotated_pole.nc'))

# Create the regridder
regridder = ESMFBilinear().regridder(global_air_temp, rotated_psl)

# Save the regridder.
save_regridder(regridder, "saved_regridder.nc")

# Load saved regridder.
loaded_regridder = load_regridder("saved_regridder.nc")

# Use loaded regridder.
result = loaded_regridder(global_air_temp)
```

## Expected behaviour

No type error encountered

## Environment 
 - OS: Linux
 - `esmf_regrid` version: 0.11.0

stephenworsley:
Closed by #431.

PR #431, opened 2024-11-06
Fix for older iris versions (<3.10)

stephenworsley:
Addresses #429.


ESadek-MO (approved):
This looks good to me, thanks @stephenworsley

stephenworsley:
For the record, the readthedocs errors seem to have been caused by this block of code
here: https://github.com/SciTools/cf-units/blob/446fcc2323c01146448366593a2766e62b1a9e82/cf_units/__init__.py#L186-L191

it seems like this might have been caused by there being no existing setting for `LC_NUMERIC` (or, somehow, the existing setting was invalid), so when cf_units attempts to revert it back to the original setting, it breaks. I'm not sure what the defaults are for readthedocs, but setting them explicitly seems to do the job.

PR #435, opened 2024-11-14
Update pyproject.toml and .pre-commit-config.yaml for common SciTools standards

stephenworsley:
Follow up issue here https://github.com/SciTools/iris-esmf-regrid/issues/437


.pre-commit-config.yaml
```diff
@@ -1,5 +1,26 @@
...
+# Alphabetised, for lack of a better order.
+files: |
+    (?x)(
+        benchmarks\/.+\.py|
+        docs\/.+\.py|
+        esmf_regrid\/.+\.py|
+        noxfile\.py|
+        pyproject\.toml|
+        setup\.py|
+        src\/.+\.py
```
HGWright, line 20: This doesn't seen necessary since we don't have a src file. Although it doesn't seem like it is a problem since its not causing errors.

pyproject.toml
```diff
@@ -69,3 +69,249 @@ docs = {file = ["requirements/optional-docs.txt"]}
...
+[tool.repo-review]
+ignore = [
+    # https://learn.scientific-python.org/development/guides/style/#PC170
+    "PC170",  # PyGrep hooks
+    # https://learn.scientific-python.org/development/guides/style/#PC180
+    "PC180",  # Uses prettier
+    "PY005",  # Test folder location
+    "GH104",  # Unique names for upload artifact
+    "GH212",  # Require GHA update grouping
+]
```
HGWright, line 82: Seems inconsistent to do the link for 2 and not all, your call if you want to remove the links that are there or add the rest.

HGWright (changes requested):
Thanks. Looks almost all good, just a couple of comments to consider.

pyproject.toml
stephenworsley: The other equivalent links don't work, I agree it does make it look a bit odd with just two though.

HGWright (approved):
Thanks. LGTM.

Issue #433, opened 2024-11-11
Cannot save regridder when providing precomputed_weights


rhyspoulton:
## 🐛 Bug Report

I am attempting to save a regridder created using the `ESMFBilinearRegridder` and providing  `precomputed_weights`, but I am enountering the following error when calling `save_regridder(regridder, 'saved_regridder.nc')`:
```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[226], line 2
      1 regridder = esmf_regrid.schemes.ESMFBilinearRegridder(src_gl_cube, tar_gl_cube, precomputed_weights = combined_weights, mdtol=1)
----> 2 save_regridder(regridder, "saved_comb_regridder.nc")
      3 print(regridder.regridder.esmf_version)

File ~/.conda/envs/iris-esmf-regrid/lib/python3.12/site-packages/esmf_regrid/experimental/io.py:270, in save_regridder(rg, filename)
    267 for cube in cube_list:
    268     cube.attributes = attributes
--> 270 iris.fileformats.netcdf.save(cube_list, filename)
...
File ~/.conda/envs/iris-esmf-regrid/lib/python3.12/site-packages/iris/fileformats/netcdf/saver.py:271, in _setncattr(variable, name, attribute)
    269 assert hasattr(variable, "THREAD_SAFE_FLAG")
    270 attribute = _bytes_if_ascii(attribute)
--> 271 return variable.setncattr(name, attribute)

File src/netCDF4/_netCDF4.pyx:3087, in netCDF4._netCDF4.Dataset.setncattr()

File src/netCDF4/_netCDF4.pyx:1888, in netCDF4._netCDF4._set_att()

TypeError: illegal data type for attribute b'ESMF_version', must be one of dict_keys(['S1', 'i1', 'u1', 'i2', 'u2', 'i4', 'u4', 'i8', 'u8', 'f4', 'f8']), got O
```
And if print out the  I print out the `emsf_version` from the regridder, I can see that is `None` which is causing the issue with saving:
```python
In  [1]: regridder = esmf_regrid.schemes.ESMFBilinearRegridder(src_cube, tar_cube, precomputed_weights = weights)

In  [2]: regridder.regridder.esmf_version
Out [2]: None
```

## How To Reproduce
Steps to reproduce the behaviour:

Call the `save_regridder` with a regridder that has `precomputed_weight` supplied e.g.:
```python
# Create the regridder
[...]

PR #462, opened 2025-02-14
fix precomputed regridder save bug

stephenworsley:
Addresses #433.


bjlittle (changes requested):
@stephenworsley This looks ideal 👍 

The only thing I'd ask you to consider is pushing the defensive `esmf_version` logic up into the regridder instance, such that it defaults to a known "unknown" string value (defined by a suitable module constant e.g., `ESMF_VERSION_DEFAULT: str = "unknown"`.

You can then control setting the `esmf_version` to an actual `esmf` version or nullify it for appropriate operations (like the user providing pre-computed weights). But in essence, it's always a string and never `None`. 

So `esmf_version` will always be a string, thus loading/saving should always "just work"™️.

Perhaps also consider (not necessarily in this PR) what matadata on a regridder is read-only or mutable.

Anyways, have a ponder and I'm happy to go with whatever you recommend 💯

esmf_regrid/tests/unit/experimental/io/test_round_tripping.py
```diff
@@ -548,3 +548,18 @@ def test_generic_regridder_resolution(tmp_path, scheme):
...
+
+def test_precomputed(tmp_path):
+    """Test save/load round tripping for regridders made with precomputed weights."""
+    original_rg, src = _make_grid_to_mesh_regridder(
+        method=None, regridder=ESMFAreaWeightedRegridder
+    )
+    tgt = original_rg._tgt
+    weights = original_rg.regridder.weight_matrix
+    original_rg = ESMFAreaWeightedRegridder(src, tgt, precomputed_weights=weights)
+    assert original_rg.regridder.esmf_version == "N/A"
```
bjlittle, line 561: @stephenworsley Can you make this `esmf_regrid.esmg_regridder.ESMF_NO_VERSION` instead?

That way it continues to work independent of the value 👍

bjlittle:
@pre-commit.ci autofix

bjlittle (approved):
@stephenworsley LGTM 🎉

PR #464, opened 2025-02-25
Add python 3.13 support, drop 3.10

stephenworsley:
Keep python versions up to date with iris.


benchmarks/asv.conf.json
```diff
@@ -23,7 +23,7 @@
     //  * No build-time environment variables.
```
pp-mo, line 23: I know it's no change, but I think the highlighting here is really complaining that "//" comments aren't really allowed in JSON syntax.

https://stackoverflow.com/questions/244777/can-comments-be-used-in-json/4183018#4183018

But the comments there are not encouraging, 
it seems if you want to be 100% "correct" there is no decent way around this.
So if this is useful, and we are currently OK with it, it should probably stay

pp-mo (approved):

pp-mo (approved):
I assume the continuing test fails are expected ??

stephenworsley:
@pp-mo  I'm not entirely sure why the tests are failing here, looks like its something to do with not being able to hash the files in requirements properly. I forget how we expect the CI to behave when updating python, I know it's often not pretty but I would have expected it to do better than this.

bjlittle (approved):
@stephenworsley `py313` happy days 😎

PR #490, opened 2025-03-26
Use new inbuilt ASV benchmark skipper

trexfeathers:
Noticed this while setting up the official templated benchmark files.


trexfeathers:
During review we should check that the CI still successfully skipped the benchmarks in question.

PR #499, opened 2025-04-14
fix unit test schemes dunder imports

bjlittle:
Kinda surprised this ever worked.

To be fair, `mypy` was kinda-sorta compliaining about it in a non-obvious way, but I think this explains the inclusion of the `no_namespace_packages` option being enabled ... which is no longer the case.


stephenworsley (approved):
Looks good.

PR #500, opened 2025-04-14
Modify shared benchmarking ops to allow optional failure when comparing commits

trexfeathers:
I believe we've been missing all regressions since #361.

### DIFF
--- a/benchmarks/bm_runner.py
+++ b/benchmarks/bm_runner.py
@@ -100,7 +100,7 @@ def _setup_common() -> None:
     _echo("Setup complete.")
 
 
-def _asv_compare(*commits: str) -> None:
+def _asv_compare(*commits: str, fail: bool = False) -> None:
     """Run through a list of commits comparing each one to the next."""
     _commits = [commit[:8] for commit in commits]
     for i in range(len(_commits) - 1):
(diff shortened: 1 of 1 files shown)


trexfeathers:
# What the test commits show

| commit | description | GHA | what this demonstrates |
| --- | --- | --- | --- |
| deddb11 | The proposed change | [`40513851523`](https://github.com/SciTools/iris-esmf-regrid/actions/runs/14448165300/job/40513851523) | The proposed change does NOT cause CI failures when no performance regressions are found |
| ecf8ee9 | introducing a false regression | [`40516320872`](https://github.com/SciTools/iris-esmf-regrid/actions/runs/14448895320/job/40516320872) | The proposed change will cause a CI failure when a performance regression is found |
| 7f1a3fe | removing the proposed code change | [`40574624613`](https://github.com/SciTools/iris-esmf-regrid/actions/runs/14467949918/job/40574624613) [^1] | Without the proposed change, the CI will pass even when regressions are found - i.e. in the current status quo we are not noticing performance regressions e.g. when Dask changes |

[^1]: This is actually the GHA run from 993f9d9 - I had to resolve conflicts with `main` before anything would run

stephenworsley (approved):
Looks good. Definitely worth getting this in ASAP.

PR #505, opened 2025-04-14
use groups for dependabot

bjlittle:
Reference, see:

- [groups](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--)
- [commit-message](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#commit-message--)


stephenworsley (approved):
Looks good.

PR #506, opened 2025-04-14
enable dependabot pypi package version updates

bjlittle:
This pull-request enables `dependabot` updates for PyPI package version updates, akin to our lockfile updates but using `dependabot` instead. 

Reference, see:

- [package-ecosystem](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#package-ecosystem-)
- [groups](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--)
- [commit-message](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#commit-message--)


stephenworsley (changes requested):
Looks good, but given that this implies the introduction of minimum pins in the requirements I think a Changelog entry would be good. It's also worth considering if introducing and controlling max pins in the .yml should also happen, if it ought to be introduced simultaneously or if that is a seperable issue.

bjlittle:
@stephenworsley I've introduced the capping pins for PyPI package versions, which will be explicitly updated automatically by `dependabot` via pull-request.

I'll also add a changelog entry.

stephenworsley (approved):
Looks good.

PR #508, opened 2025-04-15
Replace asserts

stephenworsley:


bjlittle (changes requested):
@stephenworsley This looks good to me.

One question though ... did you also want to bank test coverage for these exceptions?

I guess this is beyond the scope of what you had before otherwise you would have had failing unit tests due to the migration from `assert` statements.

Happy to go with whatever you prefer 👍

bjlittle (approved):

PR #516, opened 2025-04-25
add nodefaults to yml

bjlittle:


ESadek-MO (approved):
from @SciTools/peloton: Thanks @bjlittle, looks good to me!

PR #517, opened 2025-04-25
Update IER benchmarking to align with templating

trexfeathers:
Now we've create templates for benchmarking files (SciTools/.github#119), there are deviations between Iris and the templates. This PR removes those deviations.


benchmarks/benchmarks/generate_data/__init__.py
trexfeathers, line 1: In case it wasn't obvious, this was originally `generate_data.py`, and now it's a two-file directory (`__init__` and `cubes`).

Done primarily to help with templating - the template has directory structure.

benchmarks/asv_delegated_conda.py
trexfeathers, line 1: Iris abandoned this approach last year, with the newer approach being the one in the template. If you're interested in more details you can have a look at SciTools/iris#5963.

benchmarks/asv.conf.json
```diff
@@ -3,30 +3,25 @@
...
-    //  * No build-time environment variables.
-    //  * Is run in the same environment as the ASV install itself.
-    "delegated_env_commands": [
-        "PY_VER=3.12 nox --envdir={conf_dir}/.asv/env/nox01 --session=tests --install-only --no-error-on-external-run --verbose"
+        "We do however need to install the custom benchmarks for them to be",
+        "usable."
     ],
-    // The parent directory of the above environment.
-    //  The most recently modified environment in the directory will be used.
-    "delegated_env_parent": "{conf_dir}/.asv/env/nox01"
```
trexfeathers, line 31: SciTools/iris#6329 provides the original explanation for moving away from this concept. The 6329 changes were adopted into the template.

benchmarks/bm_runner.py
```diff
@@ -22,13 +27,13 @@
 GH_REPORT_DIR = ROOT_DIR.joinpath(".github", "workflows", "benchmark_reports")
 
 # Common ASV arguments for all run_types except `custom`.
-ASV_HARNESS = "run {posargs} --attribute rounds=4 --interleave-rounds --show-stderr"
+ASV_HARNESS = "run {posargs} --attribute rounds=3 --interleave-rounds --show-stderr"
```
stephenworsley, line 30: I'm curious why `rounds` ended up changed here, I assume that this value is somewhat arbitrary and this at least keeps it cosistent with the rest of SciTools. I don't think this change is a problem, but I'd be interested if there was a motivation for it.

stephenworsley (approved):
Looks good. Seems consistent with the template and the iris-esmf-regrid specific bits make sense.

benchmarks/bm_runner.py
trexfeathers: `rounds=4` came from Iris originally. SciTools/iris#5949 and SciTools/iris#5981 made Iris' benchmarks accurate enough to reduce the number of rounds, without causing a load of false positive results (SciTools/iris#6234). So `rounds=3` is what ended up in the template.

I don't consider Iris-esmf-regrid to have any extra vulnerability to noise, so aligning with the template seemed sensible.

PR #558, opened 2025-07-28
in_dtype is out_dtype for Nearest Regridding

### LINKED ISSUE #481: Nearest neighbour regrid results in a change of data type
## 🐛 Bug Report

When regridding int32 data with iris-esmf-regrid, the datatype of the output is changed to float64.  For nearest neighbour regrid, I think the source data type should be preserved - the data in an output cell should be the same as the data in the nearest source cell.  To my mind, that includes preserving the original source data type.

I do think the conversion from integer to floating point does make sense for area weighted or linear regrid.  But nearest neighbour is not the same.  There's no interpolation per se, instead data is mapped from the source grid to the target.

## How To Reproduce
Steps to reproduce the behaviour:

Adapted from [this example](https://iris-esmf-regrid.readthedocs.io/en/latest/userguide/examples.html#simple-regridding) (and using the same data files):

```
>>> import iris
>>> import numpy as np
>>> from esmf_regrid.schemes import ESMFNearest
>>> source_mesh_cube = iris.load_cube("mesh_cube.nc")
>>> source_mesh_cube.data = np.int32(source_mesh_cube.data)
/data/apps/sss/environments/default-2024_11_26/lib/python3.12/site-packages/numpy/ma/core.py:467: RuntimeWarning: invalid value encountered in cast
  fill_value = np.array(fill_value, copy=False, dtype=ndtype)
>>> source_mesh_cube.dtype
dtype('int32')
>>> target_grid_cube = iris.load_cube("grid_cube.nc")
>>> result = source_mesh_cube.regrid(target_grid_cube, ESMFNearest())
>>> result.dtype
dtype('float64')
>>> import esmf_regrid
>>> esmf_regrid.__version__
'0.11.0'
```

## Expected behaviour
I would expect the result datatype to be the same as the source data type.

## Environment 
 - OS and version: RHEL 9.5
 - `esmf_regrid` version: 0.11.0

HGWright:
This PR closes #481. Nearest neighbour regridding now outputs the same dtype as the is the input dtype.

### DIFF
--- a/src/esmf_regrid/esmf_regridder.py
+++ b/src/esmf_regrid/esmf_regridder.py
@@ -3,6 +3,7 @@
 import numpy as np
 from numpy import ma
 import scipy.sparse
+from scipy.sparse import safely_cast_index_arrays
 
 import esmf_regrid
 from esmf_regrid import Constants, check_method, check_norm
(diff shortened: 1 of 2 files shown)


HGWright:
pre-commit.ci autofix

hdyson:
I did raise on the [original issue](https://github.com/SciTools/iris-esmf-regrid/issues/481#issuecomment-2737409699) that I thought this change might need some careful documentation.  It may be surprising to the end user that the data type of their result depends on the regridding scheme used?  Do you think this should be documented, or do you think this is sufficiently obvious that it doesn't need docs?

PR #649, opened 2026-04-15
Add utility for finding cell areas

### LINKED ISSUE #645: Calculate area weights
ESMF can be used to calculate the area weights for grids and meshes, it would be a nice convenience to expose this with iris-esmf-regrid.

stephenworsley:
Closes #645.

### DIFF
--- a/src/esmf_regrid/tests/unit/util/test_find_area.py
+++ b/src/esmf_regrid/tests/unit/util/test_find_area.py
@@ -17,14 +17,23 @@ def test_find_area_mesh():
     cube = _gridlike_mesh_cube(5, 6)
     areas = find_area(cube)
     assert areas.shape == (5 * 6,)
+    # Check that the total area is equal to the area of the unit sphere, i.e. pi * 4
     assert np.allclose(areas.sum(), np.pi * 4)
 
 
 def test_find_area_grid():
     """Test find_area function on a rectilinear grid."""
+    # Generate a grid with regularly spaced latitudes and longitudes.
     cube = _grid_cube(5, 6, (-180, 180), (-90, 90), circular=True)
     areas = find_area(cube)
     assert areas.shape == (6, 5)
+    # Check that cells between the same latitude bounds have approximately equal areas.
+    assert np.allclose(areas[:, 1:], areas[:, :-1])
+    # Check that equivalent latitude bounds above and below the equator are approximately equal.
+    assert np.allclose(areas[:3], areas[:2:-1])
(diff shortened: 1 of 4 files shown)


src/esmf_regrid/tests/unit/util/test_find_area.py
```diff
@@ -0,0 +1,44 @@
...
+
+import numpy as np
+
+from esmf_regrid.tests.unit.schemes.test__cube_to_GridInfo import (
+    _curvilinear_cube,
+    _grid_cube,
+)
+from esmf_regrid.tests.unit.schemes.test__mesh_to_MeshInfo import (
+    _gridlike_mesh_cube,
+)
```
trexfeathers, line 11: Not for this PR, but it's probably time to start thinking about an `esmf_regrid.tests.stock` module!

src/esmf_regrid/util.py
```diff
@@ -0,0 +1,34 @@
...
+            msg = (
+                f"Area can only be calculated on mesh faces. "
+                f"Cube had location '{cube.location}'."
+            )
+            raise ValueError(msg)
+        sdo = _mesh_to_MeshInfo(cube.mesh, cube.location)
+    else:
+        sdo = _cube_to_GridInfo(cube)
+    field = sdo.make_esmf_field()
+    field.get_area()
```
trexfeathers, line 30: Thanks @stephenworsley for the link to the underlying ESMF routine:

https://earthsystemmodeling.org/docs/release/latest/ESMF_refdoc/node5.html#SECTION050367300000000000000

The below calculation is appropriate if the area is represented in square radians, which ESMF uses when the grid or mesh is on the surface of a sphere. @stephenworsley has explained that spherical is the default behaviour[^1] 👍

[^1]: Non-defaults such as cartesian could be specified using ESMPy's [`CoordSys`](https://earthsystemmodeling.org/esmpy_doc/release/latest/html/CoordSys.html#esmpy.api.constants.CoordSys) flag.

src/esmf_regrid/tests/unit/util/test_find_area.py
trexfeathers, line 1: So far all these tests use a full sphere, which makes for simple, readable tests! Should we have one that can prove it works on a region, too? Not sure how you generate a region/parent-sphere that will provide a known correct answer...

trexfeathers (changes requested):
Nice, this makes sense @stephenworsley. A great thing to capture as a convenience function. I just have a single request, regarding the testing.

trexfeathers (approved):

PR #655, opened 2026-05-12
Make regridding serializable

stephenworsley:
The lazy regridding operation requires dask to keep hold of a `Regridder` object. This in turn keeps hold of a cartopy `crs` object. This object can cause errors when the dask scheduler is not "threads". This is fixed by creating a minimal version of the regridder object which does not hold on to these objects.


stephenworsley:
Benchmarks are actually improving, asv currently treats all changes as an error.

pp-mo (approved):
OK, I checked this fairly carefully, and it all seems to make now.

PR #668, opened 2026-06-26
Add trusted publishing for pypi and test-pypi

HGWright:
Same as [python-stratify#425](https://github.com/SciTools/python-stratify/pull/425)

Requires configuration on PyPi and Test-PyPi


stephenworsley (approved):
Looks good.

